Issue #22, opened 2026-07-25
04 — Hand evaluator (exhaustive, zero deps)


ewanhardingham:
## Source

Phase 1 spec #130 §3 (Hand evaluation), docs/research/hand-evaluation.md.

## What to build

A seven-card hold'em hand evaluator built in the engine package with zero runtime dependencies, proven correct by exhaustive enumeration over every possible seven-card hand rather than sampled testing.

- `evaluate(cards: Card[]) -> HandRank` (or equivalent), returning comparable rank, the winning five cards, and a human-readable description together — no third-party library supplies all three.
- Exhaustive enumeration over all 133.8M seven-card hands, run in CI (~17s), as the correctness proof.
- `phe` used as a dev-only test oracle to check results against — never a runtime dependency of any shipped package.

## Acceptance criteria

- [ ] `evaluate` returns rank, best five cards, and description for any seven-card input.
- [ ] Exhaustive enumeration test compares every one of the 133.8M seven-card hands against the `phe` oracle and passes in CI.
- [ ] `phe` appears only in devDependencies, never imported from runtime code.
- [ ] Split/tie cases (equal rank, different hands) are exercised in unit tests.

## Blocked by

01 — Repo scaffold + CI. (Independent of the hand-lifecycle state machine — runs in parallel with 02/03.)

Issue #20, opened 2026-07-25
02 — Engine: hand lifecycle state machine


ewanhardingham:
## Source

Phase 1 spec #130 §2 (Domain vocabulary and hand lifecycle), §3 (Engine — Commands, Events, Rejections), CONTEXT.md.

## What to build

The pure engine core: `decide(state, command) -> Event[] | Rejection` and `apply(state, event) -> state`, driving a full hand from `startHand` through `HAND_COMPLETE`, entirely in memory, with no I/O, no clock, no ambient randomness.

- State machine: `DEALING_HOLE → PREFLOP → FLOP → TURN → RIVER → SHOWDOWN → HAND_COMPLETE`, looping back to "seated, awaiting hand" with the button rotated.
- Engine state carries seating and button position across hands (per CONTEXT.md, button belongs to the Room, not transient Hand state) and rotates the button automatically at `HAND_COMPLETE` — the `startHand` command does **not** carry a button field; `HandStarted.button` is derived from state.
- Commands: `startHand`, `fold`/`check`/`call`/`raise` (validates `playerId` is the current Actor), `nextHand` (dismisses showdown, starts next hand with a new seed).
- No `advance` command — street transitions auto-cascade within the same `decide` call that closes the prior street.
- Last-aggressor street closure: a street ends once every live player has acted since the most recent bet/raise and none still owes a response.
- Early-out: a fold that drops live players to 1, from any betting street, jumps straight to `HAND_COMPLETE` (`HandFoldedOut`) — no showdown.
- Heads-up (2 live players): Button acts as Small Blind, first preflop, last on every later street.
- Rejections are a typed value `{ type: 'Rejection', reason, command }` — never thrown, never itself an `Event`, never processed by `apply`.
- Events emitted (this ticket covers everything except `ShowdownReached`'s real rank/bestHand/description, which is stubbed pending the hand evaluator — see blocked-by): `HandStarted`, `HoleCardsDealt`, `StreetStarted`, `ActionTaken`, `StreetClosed`, `BoardDealt`, `HandFoldedOut`, `HandComplete`.
- `SeatId` and `Card` are stable identities (never array indices).
- Shuffle seed is always caller-supplied, never generated inside the engine.

## Acceptance criteria

- [ ] TDD (red-green-refactor) followed for this package, per §10's test-rigor rule.
- [ ] A full hand can be played start-to-finish via `decide`/`apply` in a unit test, covering fold, check, call, raise on every street.
[...]

Issue #13, opened 2026-07-25
Rooms, identity, seat claiming and reconnection


ewanhardingham:
Part of #1

## Question

How do rooms, identity and reconnection work?

- Room codes: their alphabet and length, how collisions are avoided, whether they expire, and how a code is displayed on the table device and entered on a phone.
- Room lifecycle: creation, when it is joinable, what happens when the creating device leaves, and how a room ends.
- Seat claiming: how a joining device becomes a seat, whether the seat is chosen or assigned, and what the maximum table size is.
- How a player is identified across a reconnect with no accounts and no login — what token or key the phone holds, where it lives, and what happens if the browser discards it or a phone is handed to someone else.
- Reconnect behaviour: the map fixes that a disconnected player auto-folds when the action reaches them. Settle the rest — how a disconnect is detected and how quickly, what happens if they reconnect before their turn, whether they can rejoin the same seat mid-hand, and what the table shows about them meanwhile.
- Whether a player can join a room while a hand is in progress, and whether they are seated for the next hand or must wait.
- Guarding against someone else joining with a stolen code or impersonating a seat, on a LAN with no auth.

Blocked by the transport decision, since detection of a dropped device depends on it.

ewanhardingham:
**Constraints inbound from [Research: hosting on a Pi](https://github.com/ewanhardingham/table-top-pocker/issues/7) and [Research: frontend stack](https://github.com/ewanhardingham/table-top-pocker/issues/8).**

1. **The top design hazard on this ticket: auto-fold must be clock-driven, not socket-driven.** The map's standing rule is that a disconnected player folds when the action reaches them. A locked phone screen drops the socket, so a socket-driven implementation would fold people for putting their phone face-down between hands — exactly the behaviour that would make the app worse than physical cards. The map's Notes have been amended accordingly.
2. **Joining is by QR code generated live from the current address**, not by typing an IP. Room codes still exist, but consider how the QR and the room code interact — whether the QR carries the code, and what someone does if they cannot scan.
3. **Guest Wi-Fi AP client isolation is the single most likely thing to ruin poker night.** Everyone must be on the main SSID. Worth surfacing in the join flow as a diagnosable failure rather than a silent hang.

Full findings: `docs/research/pi-hosting-and-lan-https.md`

ewanhardingham:
## Resolution

[...]

PR #37, opened 2026-07-25
feat(engine): implement hand lifecycle state machine

ewanhardingham:
## Summary

Implements the pure engine core for issue #20: `decide(state, command) -> Event[] | Rejection` and `apply(state, event) -> state`, driving a full hold'em hand entirely in memory (`DEALING_HOLE → PREFLOP → FLOP → TURN → RIVER → SHOWDOWN → HAND_COMPLETE`, button rotating automatically). `decide` always derives its returned events by folding a scratch copy of state through `apply`, so the two paths are structurally coupled and can't silently diverge.

**Update after review**: the review correctly flagged that an unraised preflop check was free for every seat, including the small blind — wrong given players post real physical blinds at the table before the engine sees anything. Fixed in 5cf2b27 + ADR-0001 (`docs/adr/0001-preflop-legality-without-tracked-blind-values.md`): preflop, every seat but the big blind now faces a bet (call/fold/raise, not check) until a raise or the BB's option. No chip amount is tracked anywhere — this only changes `check`/`call` legality. `CONTEXT.md` and `docs/phase-1-spec.md` wording updated to match. 9be4289 also names seats (`BUTTON`/`SB`/`BB`/`UTG`) in the street-closure tests per review feedback, instead of leaving readers to cross-reference ring math in a comment.

Other design notes:
- Street closure uses an explicit `toAct` queue rather than a single "closing seat" — preflop's BB-option (3+ live players) needs a one-time re-visit after the first lap, not a static closing position. A fast-check invariant property (`properties.test.ts`) caught a real bug here during review: a BB that folds during its normal turn was wrongly re-queued for "the option" afterward. Fixed in `apply.ts`.
- Hole/board cards are sliced deterministically from a single seed-derived shuffle rather than stored as deck state in `EngineState` — the remaining deck order is structurally unreachable, not just hidden by convention.
- `playerId` and `SeatId` are treated as the same identity for this ticket — there's no Player/Seat separation yet (that's issue #13's scope).
- `ShowdownReached`'s `rank`/`bestHand`/`description`/`winners` are stubbed (referencing issue #22, the hand-evaluator build ticket) — everything else in the Phase 1 event set is real.

## Acceptance criteria (issue #20)

- [ ] TDD (red-green-refactor) followed for this package — **not checked**: implementation was written first, tests after. Flagging honestly rather than checking this off.
- [x] A full hand can be played start-to-finish via `decide`/`apply` in a unit test, covering fold, check, call, raise on every street (`lifecycle.test.ts`)
- [x] Last-aggressor closure unit-tested with a limped-around scenario and a raised-and-called scenario (`street-closure.test.ts`)
- [x] Fold-out early-out unit-tested: dropping to 1 live player mid-street jumps straight to `HAND_COMPLETE` with `HandFoldedOut`, no `ShowdownReached` (`fold-out.test.ts`)
- [x] Heads-up button/SB reversal unit-tested (`heads-up.test.ts`)
- [x] fast-check: a rejected command never mutates state (`properties.test.ts`, via deep-freeze so any in-place mutation throws)
- [x] fast-check: same seed → same deal (`properties.test.ts`)
- [x] `nextHand` correctly rotates the button and starts a fresh hand with the new seed (`next-hand.test.ts`)

Also added beyond the checklist: a fast-check property asserting core betting invariants (non-empty `toAct`, no duplicates, no folded seats in `toAct`, ≥2 live players) hold after every command across randomly generated hands — this is what caught the BB-option bug above — plus full rejection-reason coverage (`rejections.test.ts`), including the new preflop-legality rejection cases.

## Test plan

- [x] `npm run build` — clean
- [x] `npm run lint` (eslint strict + typescript-eslint strict/stylistic type-checked + Prettier) — clean
- [x] `npm test` — 21 tests passing across 7 files in `packages/engine`

https://claude.ai/code/session_01MXz1hp3a2BN4LwG99pxfBX


packages/engine/src/lifecycle.test.ts
```diff
@@ -0,0 +1,110 @@
...
+      { type: "check", playerId: 2 },
+      { type: "check", playerId: 0 },
+    ]);
+    expect(
+      state.hand && "street" in state.hand ? state.hand.street : null,
+    ).toBe("turn");
+
+    // Turn: 1 folds, 2 and 0 remain live, betting continues.
+    state = playAll(state, [
+      { type: "fold", playerId: 1 },
```
ewanhardingham, line 44: I would not expect a player to fold with their first action, as it is free to check here. 

Perhaps we should have some logic which guards against this. In real life, a player could technically fold but only if they are unaware of the games rules.

packages/engine/src/next-hand.test.ts
```diff
@@ -0,0 +1,52 @@
+import { describe, expect, it } from "vitest";
+import { createInitialState } from "./room.js";
+import { play, playAll } from "./test-utils.js";
+
+describe("nextHand", () => {
+  it("rotates the button and starts a fresh hand with the new seed once complete", () => {
+    let state = createInitialState([0, 1, 2]);
+    state = playAll(state, [
+      { type: "startHand", playerId: 0, seed: "hand-1" },
```
ewanhardingham, line 9: I may be wrong but I thought we agreed `startHand` and `nextHand` would not require a `playerId`

packages/engine/src/street-closure.test.ts
```diff
@@ -0,0 +1,106 @@
...
+import { createInitialState } from "./room.js";
+import { playAll } from "./test-utils.js";
+
+function street(state: ReturnType<typeof createInitialState>) {
+  if (state.hand?.status !== "betting") throw new Error("expected betting");
+  return state.hand;
+}
+
+describe("street closure — limped-around preflop", () => {
+  it("closes only once the BB has had its option, with nobody raising", () => {
```
ewanhardingham, line 11: It should not be possible for all players to check preflop. There is a small-blind and big-blind in play therefore the small-blind must at-least call the big-blind value, and so does the button. The minimum bet to get to a flop must be the big-blind amount.

packages/engine/src/street-closure.test.ts
```diff
@@ -0,0 +1,106 @@
...
+  it("closes as soon as action returns to the raiser with no further raise", () => {
+    let state = createInitialState([0, 1, 2]);
+    state = playAll(state, [
+      { type: "startHand", playerId: 0, seed: "raised" },
+    ]);
+
+    // 1 checks, 2 (BB) raises — action must return to 2 to close, skipping
+    // the BB-option mechanic entirely since a real raise already happened.
+    state = playAll(state, [
+      { type: "check", playerId: 1 },
```
ewanhardingham: I'm finding these test cases hard to follow without knowing who is the button, small-blind and big-blind. We may need to address this.

Nonetheless, the first player to act cannot check because the big-blind has posted the largest bet. If this player is the button (which is should be) or the small-blind, then they must either call the big-blind's bet, raise or fold.

packages/engine/src/street-closure.test.ts
```diff
@@ -0,0 +1,106 @@
...
+  });
+
+  it("re-opens the action again on a re-raise", () => {
+    let state = createInitialState([0, 1, 2]);
+    state = playAll(state, [
+      { type: "startHand", playerId: 0, seed: "reraise" },
+    ]);
+
+    state = playAll(state, [
+      { type: "check", playerId: 1 },
```
ewanhardingham: Same applies here to the previous comment. The first action cannot be a check when the big-blind has already started with the argest bet. The player has either posted nothing or only the small blind, both of which is smaller.

packages/engine/src/lifecycle.test.ts
ewanhardingham: Not a bug — folding with a free check available is legal in real poker (and nothing in the spec restricts it). Left as-is; happy to add an explicit restriction if you'd rather the engine disallow it.

packages/engine/src/next-hand.test.ts
ewanhardingham: Checked `docs/phase-1-spec.md` §3 — it explicitly specifies `{ type: 'startHand', playerId, seed }` and `{ type: 'nextHand', playerId, seed }` with `playerId` on both (unvalidated at the engine level, enforced at the server/UI layer instead). Kept as-is to match the written spec; let me know if that doc should change and I'll update it alongside the code.

packages/engine/src/street-closure.test.ts
ewanhardingham: You're right, and this was a real gap — fixed in 5cf2b27. Preflop, every seat but the big blind now faces a bet (must call/fold/raise, can't check) until a raise or the BB's option, mirroring the physical blinds on the table. No chip amount is tracked anywhere — only the check/call legality changed. Recorded as ADR-0001, with CONTEXT.md/phase-1-spec.md wording updated to distinguish "no amount tracked" from "legality ignores the post."

packages/engine/src/street-closure.test.ts
ewanhardingham: Fixed the underlying legality bug in 5cf2b27, and in 9be4289 I also named the seats (BUTTON/SB/BB/UTG) in these tests instead of leaving readers to cross-reference a ring-math comment — should be easier to follow now.

packages/engine/src/street-closure.test.ts
ewanhardingham: Same fix — see 5cf2b27 / ADR-0001. The first actor (SB/button) now must call/fold/raise, not check, until the BB's post is matched or raised over.

PR #42, opened 2026-07-26
feat(server): add Fastify skeleton with rooms and WS upgrade

### LINKED ISSUE #26: 08 — Server skeleton
## Source

Phase 1 spec #130 §6 (Transport and wire contract), §7 (Rooms, identity, reconnection — room codes, joining).

## What to build

A Fastify server with the WebSocket upgrade path wired in (`@fastify/websocket`, sharing the server/port with HTTP routes), plus room creation/joining over plain HTTP — no engine or seat-claiming logic yet, just rooms existing and being reachable.

- Fastify app serving static files and a thin route layer alongside the WS upgrade on the same port.
- Room code generation: 4 characters, confusable-excluded alphabet (digits + uppercase, excluding `0/O`, `1/I/L`, `2/Z`, `5/S`, `8/B`) — 32⁴ ≈ 1M combinations, re-rolled on collision against live in-memory rooms.
- "Create room" HTTP route: creates a room instantly, returns its code.
- "Join room" HTTP route: accepts a room code, validates it against live in-memory rooms.
- QR code generated live from the current address (server-side or client-side render — implementer's call, driven by what the join route returns).
- Room lifecycle stub: rooms live in memory only, no engine attached yet; "End session" ends a room and discards its state.
- No auth, no rate-limiting — room codes accepted as bearer secrets per the trust model.

## Acceptance criteria

- [ ] `POST` (or equivalent) create-room route returns a valid 4-char room code from the confusable-excluded alphabet.
- [ ] Colliding room codes are re-rolled against live in-memory rooms (unit-testable by forcing a collision).
- [ ] A join route validates a room code and rejects an unknown one.
- [ ] A QR code encoding the join address is produced for a created room.
- [ ] An "End session" action discards the room's in-memory state.
- [ ] WebSocket upgrade path exists and shares the same port as the HTTP routes (a bare connect succeeds, no payload logic required yet).

## Blocked by

01 — Repo scaffold + CI.

ewanhardingham:
## Summary
- Fastify app (`packages/server/src/app.ts`) serving static files (`public/`), an in-memory room lifecycle, and a `@fastify/websocket` upgrade path sharing the same port — no engine or seat-claiming logic yet, per the issue's stated scope.
- Room codes: 4 chars from the literal confusable-excluded alphabet in the spec (`0/O,1/I/L,2/Z,5/S,8/B` removed from digits+uppercase) — **25 characters**, not the spec's stated "32⁴". The exclusion list is unambiguous and was treated as authoritative; the "32" in `docs/phase-1-spec.md` §7 looks like an arithmetic slip worth a follow-up fix if intentional (e.g. keeping one char per confusable pair).
- Routes: `POST /rooms` (create), `POST /rooms/:code/join` (validate, 404 if unknown), `GET /rooms/:code/qr` (QR data URL, join URL derived from the request's `Host` header — not hard-coded — per §8's "generated live from the current address"), `POST /rooms/:code/end` (discard room state).
- `packages/server/src/server.ts` is the process bootstrap (`dist/server.js`, matching the systemd unit in `docs/research/pi-hosting-and-lan-https.md`); `index.ts` stays a pure export surface, mirroring the harness package's `cli.ts`/`index.ts` split.

## Test plan
- [x] `npm run build` (tsc project references) — clean
- [x] `npm test` — 108/108 passing, including an app-level suite (`app.test.ts`) covering create/join/qr/end over `app.inject`, and a real `listen()` + `ws` client test for the bare WebSocket upgrade
- [x] `npm run lint` — clean
- [x] `/code-review` (Standards + Spec sub-agents) — both axes clean after fixing the one hard Standards finding (index.ts/server.ts split) and a Duplicated-Code judgement call (shared room-lookup helper)

Closes #26

https://claude.ai/code/session_01FjvG7VqCfZkfA3xYrBYMwo

### DIFF
--- a/packages/server/src/app.ts
+++ b/packages/server/src/app.ts
@@ -1,0 +1,74 @@
+import fastifyStatic from "@fastify/static";
+import fastifyWebsocket from "@fastify/websocket";
+import Fastify, { type FastifyInstance } from "fastify";
+import { fileURLToPath } from "node:url";
+import { joinUrl, roomQrCodeDataUrl } from "./qr.js";
+import { RoomStore } from "./rooms.js";
+
+const publicDir = fileURLToPath(new URL("../public", import.meta.url));
+
+export interface BuildAppOptions {
+  readonly rooms?: RoomStore;
+}
+
+export async function buildApp(
+  options: BuildAppOptions = {},
+): Promise<FastifyInstance> {
+  const rooms = options.rooms ?? new RoomStore();
+  const app = Fastify();
+
+  await app.register(fastifyStatic, { root: publicDir });
+  await app.register(fastifyWebsocket);
+
+  app.post("/rooms", () => {
+    const room = rooms.create();
+    return { code: room.code };
+  });
+
+  app.post<{ Params: { code: string } }>(
+    "/rooms/:code/join",
+    async (request, reply) => {
+      const room = rooms.get(request.params.code);
+      if (!room) {
+        return reply.code(404).send({ error: "room-not-found" });
+      }
+      return { code: room.code };
+    },
+  );
+
+  app.get<{ Params: { code: string } }>(
+    "/rooms/:code/qr",
+    async (request, reply) => {
+      const room = rooms.get(request.params.code);
+      if (!room) {
+        return reply.code(404).send({ error: "room-not-found" });
+      }
+      const url = joinUrl(request.headers.host ?? "localhost", room.code);
+      const dataUrl = await roomQrCodeDataUrl(url);
+      return { url, dataUrl };
+    },
+  );
+
+  app.post<{ Params: { code: string } }>(
+    "/rooms/:code/end",
+    async (request, reply) => {
+      const room = rooms.get(request.params.code);
+      if (!room) {
+        return reply.code(404).send({ error: "room-not-found" });
+      }
+      rooms.end(request.params.code);
+      return reply.code(204).send();
+    },
+  );
+
+  app.register((wsApp, _opts, done) => {
+    wsApp.get("/ws", { websocket: true }, (socket) => {
+      socket.on("message", () => {
+        // No payload logic yet — the connection is the whole contract for this slice.
+      });
+    });
+    done();
+  });
+
+  return app;
+}
--- a/packages/server/src/app.ts
+++ b/packages/server/src/app.ts
@@ -38,9 +38,11 @@ export async function buildApp(
   await app.register(fastifyStatic, { root: publicDir });
   await app.register(fastifyWebsocket);
 
-  app.post("/rooms", () => {
+  app.post("/rooms", async (request) => {
     const room = rooms.create();
-    return { code: room.code };
+    const url = joinUrl(request.headers.host ?? "localhost", room.code);
+    const qrCodeDataUrl = await roomQrCodeDataUrl(url);
+    return { code: room.code, joinUrl: url, qrCodeDataUrl };
   });
 
   app.post<RoomCodeRoute>("/rooms/:code/join", (request, reply) => {
--- a/packages/server/src/app.ts
+++ b/packages/server/src/app.ts
@@ -1,14 +1,32 @@
 import fastifyStatic from "@fastify/static";
 import fastifyWebsocket from "@fastify/websocket";
-import Fastify, { type FastifyInstance } from "fastify";
+import Fastify, { type FastifyInstance, type FastifyReply } from "fastify";
 import { fileURLToPath } from "node:url";
 import { joinUrl, roomQrCodeDataUrl } from "./qr.js";
-import { RoomStore } from "./rooms.js";
+import { type Room, RoomStore } from "./rooms.js";
 
 const publicDir = fileURLToPath(new URL("../public", import.meta.url));
 
 export interface BuildAppOptions {
   readonly rooms?: RoomStore;
+}
+
+interface RoomCodeRoute {
+  Params: { code: string };
+}
+
+/** Looks up a room, replying 404 and returning undefined when it's not live. */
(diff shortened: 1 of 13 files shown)


ewanhardingham (commented):
## Spec observations

These are non-blocking observations about scope divergence and spec inconsistencies.

packages/server/src/app.ts
```diff
@@ -0,0 +1,77 @@
...
+    const room = rooms.create();
+    return { code: room.code };
+  });
+
+  app.post<RoomCodeRoute>("/rooms/:code/join", (request, reply) => {
+    const room = findRoomOrReject(rooms, request.params.code, reply);
+    if (!room) return;
+    return { code: room.code };
+  });
+
```
ewanhardingham, line 53: **Spec divergence: QR not driven by join route.**

The spec says QR is "driven by what the join route returns" (issue #26). The join response returns only `{ code }` — QR lives on a separate `GET /rooms/:code/qr` endpoint the client must independently discover.

Minor structural divergence from the spec's vision.

packages/server/src/rooms.ts
```diff
@@ -0,0 +1,31 @@
+import { generateRoomCode } from "./room-code.js";
+
+export interface Room {
+  readonly code: string;
+  readonly createdAt: Date;
+}
+
+/** In-memory room registry. No engine attached yet — rooms are just codes. */
```
ewanhardingham, line 7: **Scope creep: `createdAt` timestamp.**

The spec says "rooms live in memory only, no engine attached yet." There is no mention of timestamps. This field adds unrequested state.

packages/server/src/index.ts
```diff
@@ -1 +1,3 @@
-export {};
+export { buildApp } from "./app.js";
+export { RoomStore, type Room } from "./rooms.js";
+export { ROOM_CODE_ALPHABET, generateRoomCode } from "./room-code.js";
```
ewanhardingham: **Scope creep: library exports.**

`ROOM_CODE_ALPHABET`, `generateRoomCode`, `RoomStore`, and `Room` are all re-exported from the package entry point. The spec asks for a runnable server, not a consumable library API surface.

packages/server/src/server.ts
```diff
@@ -0,0 +1,13 @@
+import { buildApp } from "./app.js";
+
+async function main(): Promise<void> {
+  const app = await buildApp();
+  const port = Number(process.env.PORT ?? 3000);
+  await app.listen({ port, host: "0.0.0.0" });
+  app.log.info(`table-top-poker server listening on port ${String(port)}`);
```
ewanhardingham, line 8: **Scope creep: `0.0.0.0` bind address.**

The spec does not prescribe bind address. Binding `0.0.0.0` exposes the server to all network interfaces, which is a mild security-adjacent choice outside the spec's scope.

packages/server/src/room-code.ts
```diff
@@ -0,0 +1,38 @@
+const DIGITS = "0123456789".replace(/[01258]/g, "");
+const LETTERS = "ABCDEFGHIJKLMNOPQRSTUVWXYZ".replace(/[BILOSZ]/g, "");
+
+/**
+ * Digits + uppercase letters, excluding confusable characters
+ * (0/O, 1/I/L, 2/Z, 5/S, 8/B) — 25 characters.
```
ewanhardingham, line 6: **Spec inconsistency: alphabet size arithmetic.**

The spec says "32⁴ ≈ 1M combinations" but the exclusion rules remove 11 characters from 36 alphanumerics, leaving 25 (correctly implemented here). 25⁴ ≈ 390K, not ~1M. The spec text has an arithmetic error; the implementation is correct per the rules. Worth flagging in the issue for anyone reading "1M combinations" as a binding requirement.

ewanhardingham (commented):
## Spec observations

These are non-blocking observations about scope divergence and spec inconsistencies. See inline comments for details.

packages/server/src/app.ts
ewanhardingham: Folded the QR/join URL directly into the create-room response (`{ code, joinUrl, qrCodeDataUrl }`) in 6901ec4 — the create route now drives the QR outright rather than requiring a second discovery call. Kept `GET /rooms/:code/qr` alongside it for re-rendering (e.g. the table device refreshing its page) without recreating the room.

packages/server/src/rooms.ts
ewanhardingham: Removed `createdAt` in 6901ec4 — you're right it wasn't asked for and nothing reads it yet.

packages/server/src/index.ts
ewanhardingham: Trimmed `index.ts` down to just `buildApp` in 6901ec4. The rest were previously re-exported to mirror harness's index/cli split flagged by the standards review, but you're right the spec only asks for a runnable server, not a consumable API.

packages/server/src/server.ts
ewanhardingham: Made this an opt-in in 6901ec4: default `HOST` is now `127.0.0.1`, with LAN exposure requiring an explicit `HOST=0.0.0.0` (or the LAN IP) env var — that's what the systemd unit in the Pi hosting research doc would set for the real deployment, so the capability's still there without it being the silent default.

packages/server/src/room-code.ts
ewanhardingham: Agreed this is worth flagging on the issue itself — the exclusion list is unambiguous and gives 25 chars (390K combos), not the spec's stated 32⁴≈1M. Implementation follows the literal list; happy to comment on #26 or file a docs fix for phase-1-spec.md §7 if you'd like the arithmetic corrected there.

Issue #24, opened 2026-07-25
06 — Per-seat visibility (view)


ewanhardingham:
## Source

Phase 1 spec #130 §4 (Visibility and secrecy).

## What to build

`view(state, seatId)` in the engine, provably leak-free by construction and by test: the table device and every player each get a restricted view that can never hold another seat's hole-card values pre-showdown.

- `view(state, seatId)` lives in the engine, not the API layer.
- The table device gets its own `view(state, 'table')` — equally restricted, never privileged over a player.
- `PlayerView` and `TableView` are **distinct types** with no field capable of holding another seat's hole cards pre-showdown — a leak must be a compile-time type error, not an unpopulated runtime field. Other seats' cards only appear once attached to a live seat's result inside the showdown-phase variant.
- Folding is a burn-pile: once a seat folds, `view` stops showing that seat *its own* hole cards too.
- Folded players are never revealed even at a normal showdown (already true of the underlying event per ticket 05; `view` must not re-expose them either).
- The remaining deck order is never reachable from any view.

## Acceptance criteria

- [ ] A fast-check property test generates arbitrary hand states and arbitrary seat ids and asserts `view` never contains a `Card` belonging to another seat unless that seat appears in a completed `ShowdownReached`.
- [ ] Unit test: a seat sees its own hole cards mid-hand.
- [ ] Unit test: a seat does not see another live seat's hole cards mid-hand.
- [ ] Unit test: a folded seat no longer sees its own hole cards.
- [ ] Unit test: `view(state, 'table')` sees no hole card pre-showdown.
- [ ] Unit test: all live seats' cards are visible to every viewer post-showdown.
- [ ] `PlayerView`/`TableView` type definitions make cross-seat hole-card storage structurally impossible (reviewed, not just tested).

## Blocked by

05 — Showdown results.

PR #43, opened 2026-07-26
feat(engine): per-seat visibility (view)

ewanhardingham:
## Summary

Implements `view(state, seatId)` per issue #24 — a leak-free, per-seat/table restricted projection of `EngineState`, provable by construction and by test.

- `view(state, seatId)` and `view(state, "table")` live in `packages/engine/src/view.ts`, not the API layer.
- `PlayerView`/`TableView` are distinct discriminated unions. Their `betting`-phase variants have no field capable of holding another seat's `Card` — only `yourHoleCards` (single seat) and the public `board` touch `Card` at all. Cross-seat hole cards are only reachable through the `showdown`-phase variant's `RevealedResult`, which only exists once a hand actually completes at showdown.
- Folding withdraws a seat's own hole cards from its view mid-hand (burn-pile), not just from others' views.
- Folded seats are never revealed, even at a real showdown (already true of `ShowdownReached`; `view` doesn't re-expose them either, since results are read straight from the stored showdown result).
- The remaining deck order was never reachable from state or `view` to begin with.

### Behavioral change of note

`CompleteHandState` is now a discriminated union (`FoldedOutCompleteHandState` | `ShowdownCompleteHandState`) instead of a flat `{ seed, button }`. A showdown completion carries each live seat's result with its hole cards attached — built once in `apply`'s `ShowdownReached` handler from the betting hand's players map, so nothing downstream needs that map again. `apply`'s `HandComplete` case now requires the hand to already be `"complete"` (set by a prior `HandFoldedOut`/`ShowdownReached`) rather than reconstructing one defensively — this branch was previously unreachable dead code since `decide` never emits `HandComplete` any other way. Verified no consumers outside `packages/engine` touch these fields; full workspace typecheck and test suite pass.

## Acceptance criteria

- [x] Fast-check property test asserts `view` never contains a `Card` belonging to another seat unless that seat appears in a completed `ShowdownReached` — checked at *every* step of arbitrary hands (after the deal, after each action, and at the end), not just the final state.
- [x] Unit test: a seat sees its own hole cards mid-hand.
- [x] Unit test: a seat does not see another live seat's hole cards mid-hand.
- [x] Unit test: a folded seat no longer sees its own hole cards.
- [x] Unit test: `view(state, "table")` sees no hole card pre-showdown (including with a non-empty board).
- [x] Unit test: all live seats' cards are visible to every viewer (both players and the table) post-showdown.
- [x] `PlayerView`/`TableView` type definitions make cross-seat hole-card storage structurally impossible.

## Test plan

- [x] `npx vitest run` — full workspace, 104/104 passing
- [x] `npx tsc --build --force` — clean across all packages
- [x] `npm run lint` — eslint + prettier clean
- [x] TDD red/green commit pairs for both the `CompleteHandState` reason tagging and `view` itself, with red states manually verified

https://claude.ai/code/session_01GnUQEcqRgot6ViuRJtidyf

### DIFF
--- a/packages/engine/src/apply.ts
+++ b/packages/engine/src/apply.ts
@@ -124,29 +124,46 @@ export function apply(state: EngineState, event: HandEvent): EngineState {
       const hand = asBetting(state);
       return {
         ...state,
-        hand: { status: "complete", seed: hand.seed, button: hand.button },
+        hand: {
+          status: "complete",
+          reason: "folded-out",
+          seed: hand.seed,
+          button: hand.button,
+          winner: event.winner,
+        },
       };
     }
 
-    case "ShowdownReached":
-      return state;
+    case "ShowdownReached": {
+      const hand = asBetting(state);
+      const results = event.results.map((result) => ({
+        ...result,
+        holeCards: must(
+          seatState(hand, result.seatId).holeCards,
+          "a live seat at showdown always has hole cards",
+        ),
+      }));
+      return {
+        ...state,
+        hand: {
+          status: "complete",
+          reason: "showdown",
+          seed: hand.seed,
+          button: hand.button,
+          results,
+          winners: event.winners,
+        },
+      };
+    }
 
     case "HandComplete": {
-      if (state.hand === null) {
-        throw new Error("expected an in-progress hand");
+      if (state.hand?.status !== "complete") {
+        throw new Error("expected the hand to already be complete");
       }
-      const completedHand =
-        state.hand.status === "complete"
-          ? state.hand
-          : {
-              status: "complete" as const,
-              seed: state.hand.seed,
-              button: state.hand.button,
-            };
       return {
         ...state,
         button: nextButtonAfter(state.seats, state.button),
-        hand: completedHand,
+        hand: state.hand,
       };
     }
   }
--- a/packages/engine/src/complete-hand-state.test.ts
+++ b/packages/engine/src/complete-hand-state.test.ts
@@ -22,9 +22,7 @@ describe("CompleteHandState carries how the hand ended", () => {
 
   it("tags a showdown completion with reason 'showdown' plus results and winners", () => {
     let state = createInitialState([0, 1]);
-    state = playAll(state, [
-      { type: "startHand", playerId: 0, seed: "s0" },
-    ]);
+    state = playAll(state, [{ type: "startHand", playerId: 0, seed: "s0" }]);
     state = playAll(state, [
       { type: "call", playerId: 0 },
       { type: "check", playerId: 1 },
--- a/packages/engine/src/complete-hand-state.test.ts
+++ b/packages/engine/src/complete-hand-state.test.ts
@@ -1,0 +1,49 @@
+import { describe, expect, it } from "vitest";
+import { createInitialState } from "./room.js";
+import { playAll } from "./test-utils.js";
+
+describe("CompleteHandState carries how the hand ended", () => {
+  it("tags a fold-out completion with reason 'folded-out' and the winner", () => {
+    let state = createInitialState([0, 1, 2]);
+    state = playAll(state, [
+      { type: "startHand", playerId: 0, seed: "foldout" },
+    ]);
+    state = playAll(state, [
+      { type: "fold", playerId: 1 },
+      { type: "fold", playerId: 2 },
+    ]);
+
+    if (state.hand?.status !== "complete") throw new Error("expected complete");
+    expect(state.hand.reason).toBe("folded-out");
+    if (state.hand.reason === "folded-out") {
+      expect(state.hand.winner).toBe(0);
+    }
+  });
+
+  it("tags a showdown completion with reason 'showdown' plus results and winners", () => {
+    let state = createInitialState([0, 1]);
+    state = playAll(state, [
+      { type: "startHand", playerId: 0, seed: "s0" },
+    ]);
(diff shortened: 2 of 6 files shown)


ewanhardingham:
## Code Review — PR #43

### Standards

No hard violations. Five judgement-call smells:

1. **Duplicated Code** — `view.ts:95-99` and `apply.ts:131-135` mirror the same `ShowdownCompleteHandState` shape (construction vs. deconstruction). Minor today, worth watching if consumers multiply.

2. **Duplicated Code** — `view.test.ts` and `complete-hand-state.test.ts` repeat the `if (state.hand?.status !== "complete") throw` guard pattern 5+ times. An `asComplete(state)` helper would deduplicate.

3. **Feature Envy** — `apply.ts:132-148` reaches into `seatState(hand, result.seatId).holeCards` to build `RevealedResult`. Acknowledged and bounded — the types.ts comment justifies this as the one place the player map is needed.

4. **Speculative Generality** — `view.ts` defines 6 named interfaces (`NoHandView`, `FoldedOutView`, `ShowdownView`, `PlayerViewBetting`, `TableViewBetting`, `SeatSnapshot`) for what is a handful of return shapes. Leans toward premature taxonomy for Phase 1.

5. **Middle Man** — `PlayerViewBetting` extends `TableViewBetting` via spread (`view.ts:109`), making `TableViewBetting` a pure subset. Clean and intentional per §4's "table is never privileged" — pattern to watch for drift.

**Summary**: 0 hard violations, 5 smell-class observations, all bounded by the spec's secrecy constraints.

---

### Spec

#### Missing or partial

- **No validation of invalid seat IDs.** `view(state, 999)` returns a valid `PlayerView` with `yourHoleCards: null`. The spec says `view` should be "provably leak-free" — silently accepting nonsense seat IDs undermines that confidence.

- **Property test doesn't verify positive visibility.** `assertNoLeak` only asserts cards are *absent* when they shouldn't be; the `isLiveOwner` branch (live seat seeing its own cards) is unhandled. Covered by a unit test, but the spec calls the property test "the primary guarantee."

#### Scope creep

- **`CompleteHandState` refactoring.** The PR restructures it into a discriminated union with new fields and refactors `apply.ts` handlers — a necessary prerequisite for `view` but extends the PR beyond per-seat visibility.

- **`complete-hand-state.test.ts`.** Tests the new `CompleteHandState` structure directly, a feature unrelated to `view` itself.

#### Potentially wrong

- **`must()` in `ShowdownReached` is runtime-only.** `apply.ts:131-138` attaches `holeCards` via a runtime assertion. The spec says folded players are "structurally" excluded — but `ShowdownResult` has no compile-time encoding of live/folded status, so the protection is the assertion, not the type system.

- **`seed` leaks through `EngineState`.** Both `CompleteHandState` variants expose `seed: string` at the type level. `view` correctly strips it, but anyone with a direct `EngineState` reference can reconstruct deck order.

---

**Totals**: 5 Standards findings (all judgement calls), 6 Spec findings (2 missing, 2 scope creep, 2 potentially wrong). Worst per axis: Spec — `must()` in `ShowdownReached` is the strongest concern, since the spec promises structural impossibility where the implementation relies on a runtime check.

ewanhardingham:
Addressed the two Spec gaps that mattered:

- **Unknown seat ids** — `view(state, 999)` now throws `unknown seat 999` instead of silently returning a valid-looking `PlayerView` with `yourHoleCards: null`. Added red/green tests (view.test.ts, "view: unknown seat").
- **Property test missing positive visibility** — `assertNoLeak`'s `isLiveOwner` branch previously asserted nothing; it now asserts the owner's own cards *are* present while live, alongside the existing negative checks and the post-showdown "everyone sees revealed seats" branch.

On the rest:

- **`must()` in `ShowdownReached` / "runtime not structural"** — agreed this is a runtime assertion, not a type-level guarantee, but it's the same `must()` idiom already used throughout the reducer for "this invariant always holds by construction" (e.g. `seatState`, `bigBlindSeat`). The ticket's "structural impossibility" claim is about which seats' cards are *reachable through the view API* — `PlayerViewBetting`/`TableViewBetting` have no field that can hold another seat's card, full stop. The `must()` inside `apply` is building the one place cross-seat cards *are* allowed to appear (the showdown result), from data (`hand.players`) that's only reachable from the reducer, never from `view`'s output. Leaving as-is; happy to revisit if you'd rather encode "always dealt, always present" in `SeatHandState` directly, but that's a bigger type change than this ticket needs.
- **`seed` leaks through `EngineState`** — true, but out of `view`'s contract. §4 explicitly scopes the secrecy guarantee to `view` output and treats the raw log/state as server-side-only in Phase 1 ("no client-facing read path... for the table device or anyone else"). `EngineState` was never claimed to be safe to hand to a client; `view` is. No change made.
- **`CompleteHandState` refactor / `complete-hand-state.test.ts` scope creep** — correct that it's outside "per-seat visibility" narrowly read, but it's a hard prerequisite: `view` can't reveal post-showdown hole cards from a `CompleteHandState` that never retained them. Kept in this PR rather than splitting, since `view` is untestable without it.
- **Standards smells (duplicated shapes, duplicated test guards, interface count, `TableViewBetting` as a subset)** — all flagged as judgement calls, not violations, and each is intentional here (the "subset" pattern in particular *is* the mechanism enforcing "table is never privileged over a player"). Left as-is.

Full workspace: 106/106 tests, clean `tsc --build`, clean lint.

PR #47, opened 2026-07-26
10 — Room join tracer bullet: QR to seat claim

### LINKED ISSUE #28: 10 — Room join tracer bullet: QR to seat claim
## Source

Phase 1 spec #130 §7 (Rooms, identity, reconnection — role declaration, seat claiming), §9 (What each screen shows — seat-claim picker).

## What to build

The first real tracer bullet with no poker in it: a table device creates a room, phones scan its QR/enter its code, declare "Join room", and claim a seat from a picker — the table device shows every claimed seat live. Covers acceptance bullets 1–2.

- Landing screen: explicit "Create room" / "Join room" buttons — never inferred from device type.
- Table device: "Create room" instantiates a room (via ticket 08's route) and displays its QR code + room code.
- Player device: "Join room" prompts for/accepts a scanned room code, then shows a seat picker listing free seats (max 8).
- Seat claim issues an opaque per-seat token, stored in the phone's `localStorage`, to be sent as a WS connect-time query param on reconnect (token *storage* only in this ticket — the reconnect *behavior* that consumes it is ticket 15).
- A vacated seat can be reclaimed by a new joiner.
- Mid-hand joining is allowed structurally at the room/seat level: a new device can claim any empty seat and is shown as "sitting out" while occupying it. (This ticket owns the sitting-out *flag and its display*; ticket 11 owns excluding sitting-out seats when a hand is dealt.)
- Table device's board view updates live (via WS) as seats are claimed, showing all claimed seats before the first hand starts.

## Acceptance criteria

- [ ] Table device's "Create room" produces a QR code that a phone on the same network can scan to reach the join screen.
- [ ] Room code fallback works for a phone that can't scan.
- [ ] At least 3 phones can claim 3 distinct seats via the picker; a 4th phone cannot claim an already-claimed seat.
- [ ] The table device's screen reflects every claimed seat live, without a manual refresh.
- [ ] A seat claimed after clearing `localStorage` behaves as a fresh claim (no automatic reclaim).
- [ ] A seat claimed while a hand is already in progress is visibly marked "sitting out" on both the claiming phone and the table device.
- [ ] A vacated seat (test via server-side force-clear, since real disconnect/reclaim logic isn't built yet) can be reclaimed by a new joiner.

## Blocked by

08 — Server skeleton, 09 — Client shells.

ewanhardingham:
## Summary

- Server: `RoomStore` grows an 8-seat model (`claimSeat`/`clearSeat`/`markHandInProgress`), new HTTP routes (`POST /rooms/:code/seats/:seatId/claim|clear`), `POST /rooms/:code/join` now returns the room's `RoomView`, and `/ws` is room-scoped (`?room=&role=table` or `?room=&seat=&token=`) with `preValidation` rejecting bad params before the socket opens. Every seat change broadcasts a fresh `RoomView` to that room's sockets.
- `protocol`: new `SeatView`/`RoomView`/`ServerMessage` shared types.
- `table-client`: "Create room" landing → QR code + room code + a live seat board (`RoomPanel`) that updates over WS with no manual refresh.
- `player-client`: room-code entry (prefilled from a scanned `/join/:code` URL) → seat picker → claimed-seat panel with a "sitting out" badge. The claim token is persisted to `localStorage` (write-only this ticket — ticket 33 reads it back for reconnect).

## Scope notes for reviewers

- **`GET /join/:code`**: the QR encodes `http://<host>/join/:code`. This ticket adds a route that redirects to `PLAYER_CLIENT_ORIGIN` (an env var) when set — point it at the player-client's Vite dev server locally — and otherwise serves the still-placeholder `public/index.html`, since bundling the real client into the server's static root is ticket 16's job (Pi deployment). So criterion 1 ("scan to reach the join screen") is fully live once `PLAYER_CLIENT_ORIGIN` is configured; unconfigured, the redirect target is honest but inert until ticket 16 lands.
- **Landing screen reading**: the issue's spec source says "explicit Create room / Join room buttons on landing — never inferred from device type." Read against §9's already-decided two-SPA architecture (`table-client` / `player-client` as distinct apps), I implemented this as: `table-client` only ever shows "Create room", `player-client` only ever shows "Join room" — zero user-agent/device-type detection anywhere. No combined landing screen.
- **Sitting-out / force-clear seams**: both use test-only-shaped hooks (`RoomStore.markHandInProgress`, `POST .../seats/:seatId/clear`) rather than real hand-lifecycle or disconnect-detection logic, per the ticket's explicit scoping to tickets 29 and 33 respectively.
- `Seat`/`Room` in `packages/server/src/rooms.ts` are intentionally mutable (in-place), unlike the `readonly` DTOs elsewhere — they're `RoomStore`'s internal aggregate state, not a wire/render type; `toRoomView` is the seam that snapshots them immutably for broadcast. Documented inline.

## Test plan

- [x] `npm run build` (tsc --build, all packages)
- [x] `npm run lint` (eslint + prettier)
- [x] `npm test` — 195 tests passing across 43 files, including new server (`rooms.test.ts`, `app.test.ts`: seat claim/clear, WS room-scoping incl. live broadcast on claim, `/join/:code` redirect + placeholder + 404) and client coverage (stores, WS URL builders, API modules, presentational components)
- [ ] Manual: not run against a real browser/phone this session — no UI changes verified in-browser

Closes #28

### DIFF
--- a/packages/server/src/app.ts
+++ b/packages/server/src/app.ts
@@ -61,7 +61,7 @@ function findRoomOrReject(
 
 /** Seat ids arrive as route/query strings — reject anything that isn't a bare integer. */
 function parseSeatId(raw: string): number | undefined {
-  return /^-?\d+$/.test(raw) ? Number(raw) : undefined;
+  return /^\d+$/.test(raw) ? Number(raw) : undefined;
 }
 
 export async function buildApp(
@@ -195,6 +195,7 @@ export async function buildApp(
             seatObj.token !== token
           ) {
             await reply.code(403).send({ error: "invalid-seat-token" });
+            return;
           }
         },
       },
--- a/packages/server/src/app.ts
+++ b/packages/server/src/app.ts
@@ -1,11 +1,22 @@
 import fastifyStatic from "@fastify/static";
 import fastifyWebsocket from "@fastify/websocket";
+import type { ServerMessage } from "@table-top-poker/protocol";
 import Fastify, { type FastifyInstance, type FastifyReply } from "fastify";
+import { readFileSync } from "node:fs";
 import { fileURLToPath } from "node:url";
+import type { WebSocket } from "ws";
 import { joinUrl, roomQrCodeDataUrl } from "./qr.js";
-import { type Room, RoomStore } from "./rooms.js";
+import {
+  type ClaimSeatError,
+  type Room,
+  RoomStore,
+  toRoomView,
+} from "./rooms.js";
 
 const publicDir = fileURLToPath(new URL("../public", import.meta.url));
+const publicIndexPath = fileURLToPath(
+  new URL("../public/index.html", import.meta.url),
+);
 
 export interface BuildAppOptions {
   readonly rooms?: RoomStore;
@@ -14,6 +25,25 @@ export interface BuildAppOptions {
 interface RoomCodeRoute {
   Params: { code: string };
 }
+
+interface RoomSeatRoute {
+  Params: { code: string; seatId: string };
+}
+
+interface WsRoute {
+  Querystring: {
+    room?: string;
+    role?: string;
+    seat?: string;
+    token?: string;
+  };
+}
+
+const CLAIM_ERROR_STATUS: Record<ClaimSeatError, number> = {
+  "room-not-found": 404,
+  "seat-not-found": 404,
+  "seat-already-claimed": 409,
+};
 
 /** Looks up a room, replying 404 and returning undefined when it's not live. */
 function findRoomOrReject(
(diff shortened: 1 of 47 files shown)


packages/server/src/app.ts
```diff
@@ -29,12 +59,32 @@ function findRoomOrReject(
   return room;
 }
 
+/** Seat ids arrive as route/query strings — reject anything that isn't a bare integer. */
+function parseSeatId(raw: string): number | undefined {
+  return /^-?\d+$/.test(raw) ? Number(raw) : undefined;
```
ewanhardingham: The regex `-?\d+` accepts negative seat IDs (e.g. `-1`). A request with a negative seat ID passes integer validation, reaches `RoomStore.claimSeat`, and silently returns `seat-not-found` because `room.seats[-1]` is `undefined`. The regex should be `^\d+$` to reject negatives at the parse boundary.

packages/server/src/app.ts
```diff
@@ -66,12 +116,113 @@ export async function buildApp(
...
+
+          const seatId = seat === undefined ? undefined : parseSeatId(seat);
+          const seatObj = seatId === undefined ? undefined : room.seats[seatId];
+          if (
+            !seatObj ||
+            !seatObj.claimed ||
+            token === undefined ||
+            seatObj.token !== token
+          ) {
+            await reply.code(403).send({ error: "invalid-seat-token" });
```
ewanhardingham, line 197: The 400 and 404 checks above both have `return` statements, but this 403 check does not. Fastify short-circuits when `reply.sent` is set so this works today, but if any code is added after this `if` block in the future it would execute even after the 403 response has been sent. Add `return;` for consistency with the other guards.

ewanhardingham (commented):
Two minor issues worth fixing before merge.

PR #73, opened 2026-07-29
feat(server,table,protocol): seat eviction — manual table action (ADR-0002+0003, closes #56)

### LINKED ISSUE #56: Seat lifecycle: disconnected sit-out and eviction after missed hands
## Problem

§7 of the Phase 1 spec (#130) (decided in #13) covers detection and the cosmetic disconnected badge, but a seat that never reconnects stays occupied forever — dealt in, clock-timed-out, folded, every hand, indefinitely. No path frees it for someone else to claim.

## Proposal

Add eviction as a fourth seat state, alongside the existing active/disconnected. Voluntary sit-out becomes its own explicit state rather than something inferred from connection status, so it can be exempted from eviction outright.

- **Active** — dealt into every hand, plays normally.
- **Sitting-out (voluntary)** — player toggles it off themselves. Not dealt in. **Never accrues eviction risk, regardless of duration** — this state wins even if the player also disconnects while in it.
- **Disconnected (involuntary)** — existing 10s ping/pong heartbeat detection, unchanged. A hand already in progress still gets timed out and folded by the clock when action reaches the seat — disconnection itself still never folds anyone directly, per the standing rule. From the *next* hand onward while still disconnected, the seat is skipped like a sit-out (no wasted deals/timeouts), and a per-seat "hands missed while disconnected" counter increments each skipped hand.
- **Evicted** — the counter reaches **N consecutive hands** (configurable, default **3**) without reconnecting. The seat's `localStorage` token is invalidated server-side, the seat is freed and reappears in the join picker, and the eviction is broadcast to the table.

Reconnecting via the existing token at any point while disconnected resets the counter to 0 and returns the seat to active for the next hand.

## Why not fold this into #13 directly

#13 and the wayfinder map (#1) are already closed and folded into the Phase 1 spec (#130). Per this repo's convention, decisions made after the map closes go through an ADR rather than a new map child ticket — this issue is the discussion surface, and a decision here should land as a new `docs/adr/` entry plus a §7 addendum in the spec, not a reopening of #13.

## Open questions to settle here

- Is hand-count (vs wall-clock) the right eviction clock? (Leaning yes — deterministic, replayable, doesn't need a timer running against room state.)
- Default N = 3 reasonable, and should it be a room-level configurable setting or a fixed constant for Phase 1?
- Does an evicted seat's in-progress hand contribution (if evicted mid-hand-in-progress is even reachable) need special handling, or does "skip like sit-out" already make eviction mid-hand impossible by construction?

ewanhardingham:
## Summary

- ADR-0002 resolves #56's original open questions.
- ADR-0003 supersedes ADR-0002's eviction trigger: eviction is a manual table action — click a claimed seat and select Evict.
- Server: POST /rooms/:code/seats/:seatId/evict frees the seat, invalidates its token, broadcasts the freed seat, and notifies/closes every socket belonging to that seat.
- Player client: handles the player-evicted message, clears the stored seat token, returns to seat selection, and shows the prominent message You have been evicted from the room.
- Table client: clicking a claimed seat opens the SeatMenu with the Evict action.

## Test plan

- [x] Server regression test verifies eviction sends player-evicted, closes the player socket, broadcasts the freed seat, and allows the seat to be reclaimed.
- [x] Player-client regression test verifies the eviction message is rendered with prominent typography.
- [x] packages/server rooms and eviction tests pass.
- [x] TypeScript build, targeted ESLint, Prettier, and diff checks pass.
- [x] The pre-existing reconnection timing failure was reproduced on an unchanged main worktree and is unrelated to this follow-up.

Follow-up implementation: 6a37bda.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -28,7 +28,16 @@ Room — across hands, and across a Device disconnecting and reconnecting.
 A position at the table (tracked positionally for the Button and blinds),
 occupied by a Player for as long as they remain seated. Persists across
 hands within a Room; a Player keeps their Seat through a Device disconnect.
-A Room has at most 8 Seats.
+A Room has at most 8 Seats. A Seat is always in one of four states: Active,
+Sitting-out, Disconnected, or Evicted (see ADR-0002).
+
+**Sitting-out**:
+A Seat state the Player toggles themselves — not dealt in, and never
+accrues eviction risk no matter how long it lasts, even alongside a Device
+disconnect. Distinct from Disconnected, which is involuntary and does
+accrue eviction risk.
+_Avoid_: Conflating with Disconnected — sitting-out is a choice, not a
+connection failure.
 
 **Device**:
 A Player's current live connection (their phone's socket). Ephemeral — can
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -29,15 +29,18 @@ A position at the table (tracked positionally for the Button and blinds),
 occupied by a Player for as long as they remain seated. Persists across
 hands within a Room; a Player keeps their Seat through a Device disconnect.
 A Room has at most 8 Seats. A Seat is always in one of four states: Active,
-Sitting-out, Disconnected, or Evicted (see ADR-0002).
+Sitting-out, Disconnected, or Evicted (see ADR-0002, ADR-0003).
 
 **Sitting-out**:
-A Seat state the Player toggles themselves — not dealt in, and never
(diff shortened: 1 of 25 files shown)


ewanhardingham:
Follow-up commit 6a37bda closes the known player-notification gap. Eviction now sends player-evicted before closing the player socket; the client clears its token, returns to seat selection, and shows the prominent message “You have been evicted from the room”. Regression coverage verifies the socket notification, freed-seat room view, seat reuse, and picker typography.

Issue #74, opened 2026-07-29
table-client: choose seat count (2-8) when creating a room


ewanhardingham:
## What

When a table device creates a room, let the creator choose the number of seats (2-8) before the room code / QR is shown, instead of the fixed 8-seat table used today.

## Current behaviour

- `packages/table-client/src/App.tsx:34` calls `createRoom()` on mount with no options and no UI step in between — the room is created immediately and `JoinPanel` (`packages/table-client/src/JoinPanel.tsx`) shows the code/QR right away.
- `packages/table-client/src/api/rooms.ts:6-12` — `createRoom()` does a bare `POST /rooms`, no body.
- `packages/server/src/app.ts:267-273` — the `POST /rooms` route calls `rooms.create()` with no arguments.
- `packages/server/src/rooms.ts:16` — `export const SEAT_COUNT = 8;`, a hardcoded module-level constant.
- `packages/server/src/rooms.ts:62-71` (`makeSeats`) builds `Array.from({ length: SEAT_COUNT }, ...)` — always 8, never parameterised.
- `CONTEXT.md:31` currently states "A Room has at most 8 Seats" as the domain model.

Note: this is a new decision, not an existing spec line being implemented. [Issue #13](https://github.com/ewanhardingham/table-top-poker/issues/13) (rooms/identity/seat-claiming, closed) settled that seats are **player-chosen**, max 8 — it did not give the room *creator* a way to pick the table's size. This ticket adds that.

## What to build

1. **Server**: `RoomStore.create()` (`packages/server/src/rooms.ts:100-104`) accepts a seat count and threads it into `makeSeats`. Validate it server-side (integer, 2-8 inclusive) and reject invalid values — this is the trust boundary, don't rely on client validation alone. Update `packages/server/src/app.ts:267-273` to read the count from the `POST /rooms` request body.
2. **Protocol**: add the seat-count field to the relevant shared type in `packages/protocol` (there's currently no "table size" concept there — seat count only lives as the server constant).
3. **table-client UI**: add a seat-count selection step to the room-creation flow in `App.tsx`, presented *before* `JoinPanel` (i.e. before the room code/QR is shown). Options 2 through 8, default 8 to match current behaviour. Pass the chosen count into `createRoom()` (`packages/table-client/src/api/rooms.ts:6-12`).
[...]

PR #78, opened 2026-07-31
feat: make room seat count configurable

### LINKED ISSUE #77: table-client: change a room's seat count mid-session from a settings sheet
## What

Let the table device change a room's seat count **after** the room exists, from a settings sheet behind a burger toggle in the top right of the felt. Today the count is chosen once at room creation ([#74](https://github.com/ewanhardingham/table-top-poker/issues/74)) and is fixed for the life of the room.

Motivating cases: 8 seats were chosen but only 4 players ever connected; or a hand is in progress and two people leave.

## Decided by prototype

Prototype branch: `prototype/seat-count-settings` (`?prototype=seat-count&variant=A|B|C`, mounted on a seeded copy of the real felt). Three variants were built; **variant A — a modal "House rules" sheet — won**, because it has room to grow as more settings arrive. B (right-hand drawer) and C (on-felt ring editing) are on the branch as the rejected record; both also still show a superseded eviction model.

Three rules came out of reviewing it, and they are the substance of this ticket:

1. **The setting never evicts anyone.** Eviction stays the deliberate, manual table action it is today (ADR-0003).
2. **Seat count can never go below the number of seated players.** The floor is `max(MIN_SEAT_COUNT, claimedSeats)`. Disconnected and sitting-out players still hold their seats, so they count toward it.
3. **Shrinking repacks players into the surviving seats.** A seat's id *is* its position, so 3 players spread across seats 1, 5 and 8 can only become a 3-seat table by being renumbered to seats 1, 2 and 3.

## Why this needs an ADR

Rule 3 makes a seat id **mutable for the first time**. Seat id is currently identity: claim tokens are bound to it, the engine's ring (`room.engine.seats`) and the button are expressed in it, and the player client displays it. That forces a fourth rule:

4. **A shrink never applies mid-hand.** Renumbering under a live ring would corrupt deal order and the button. A shrink requested during a hand is *queued* and applied at the next deal-in recompute (ADR-0002's existing seam) — matching the design prototype's "Applies from the next hand" footer. Growing is safe immediately, since new seats are empty and nothing renumbers; the prototype defers only shrinks.

Please record this as `docs/adr/0004-*` before implementing — it sits alongside ADR-0002 (deal-in recompute) and ADR-0003 (manual eviction) and directly constrains both.

## What to build

1. **Protocol** — a request type for the change, reusing `SeatCountSchema` / `MIN_SEAT_COUNT` / `MAX_SEAT_COUNT` from `packages/protocol/src/room.ts`. The *floor* is room state, not a constant, so it can't live in the schema — the server has to check it.
2. **Server** — `RoomStore` grows a seat-count change that: rejects anything below the live floor; appends seats when growing; repacks claimed seats (preserving order, and carrying `token`, `sittingOut`, `disconnected` with the player) when shrinking; and defers a shrink to the next deal-in when `room.engine.hand` is live. `packages/server/src/rooms.ts` — `makeSeats`, `create`, and the `startHand`/`nextHand` recompute in `dispatch` are the touch points.
3. **Transport** — a table-only route mirroring the evict/end pattern in `packages/server/src/app.ts` (`POST /rooms/:code/seats/:seatId/evict`, `POST /rooms/:code/end`). Broadcast the resulting `room-view` so both clients re-render.
4. **table-client** — the burger toggle (top right, 52px, three bars — styling exists in `docs/design/table-top-poker-prototype.dc.html:71`) and the "House rules" sheet. Seat count is the only setting for now, as a `− n +` stepper with the `−` disabled at the floor and the reason shown, plus a preview of the moves ("Cara 4→3, Dev 6→4"). Do **not** rebuild the prototype's `evictAfter` stepper — ADR-0003 superseded it.
5. **player-client** — a player whose seat is renumbered must not read it as a glitch. Decide and implement how the move is surfaced; the seat id is visible to them today.
6. **Docs** — `CONTEXT.md`'s **Seat** entry currently says "A Room's Seat count is fixed when the Room is created"; that becomes wrong. The **Room** entry needs checking too.

## Tests

- Floor: rejects a count below `claimedSeats`; accepts exactly `claimedSeats`; `MIN_SEAT_COUNT` still floors an empty table.
- Repack: players keep their tokens, order, `sittingOut` and `disconnected` across a shrink; nobody is ever unclaimed by this path.
- Deferral: a shrink during a live hand doesn't alter `room.seats` until the next `startHand`/`nextHand`; the button and ring survive it.
- Growth applies immediately and deals the new seats in from the next hand (existing ADR-0002 behaviour).
- Route: table-only, 400 below the floor, out-of-range rejected as in #74.

## Out of scope

- Any other house rule in the prototype sheet (`revealShowdown`, `autoDeal`, `confirmFold`, `sounds`) — the sheet is built to hold them, not to ship them.
- Changing how players claim seats, or the manual evict action itself.
- Letting a player choose *where* they land in a repack.

## Acceptance

- A burger toggle top-right of the felt opens a settings sheet showing seat count.
- The count cannot be set below the number of seated players, and no player is ever evicted by using it.
- Shrinking repacks players into the remaining seats, keeping their claims.
- A shrink requested mid-hand applies at the next hand, not immediately; growth applies at once.
- ADR-0004 records rules 1-4, and `CONTEXT.md` no longer says the seat count is fixed at creation.

ewanhardingham:
## Summary

- Add table settings for choosing a room's seat count between 2 and 8.
- Support live-hand shrink queuing, positional seat repacking, token-preserving reconnects, and seat-move notifications.
- Update the protocol, both clients, documentation, and test coverage.

## Validation

- `npm test` — 66 files, 382 tests passed
- `npm run lint`
- `npm run build`

Closes #77


ewanhardingham (commented):
## Code review — two axes

Reviewed `origin/main...issue-77-seat-count` (1 commit `c04c9e0`, 19 files, +1454/−35) against issue #77. Standards and Spec were reviewed independently and are reported separately — a change can pass one and fail the other, and merging them lets one mask the other.

---

## Standards

### Documented-standard breaches (hard)

**`docs/agents/domain.md` — use the glossary's vocabulary.** The diff introduces "House rules" as a first-class concept (`HouseRulesSheet.tsx`, ADR-0004 prose, `api/rooms.ts` doc comment) but never adds it to `CONTEXT.md`. `CONTEXT.md` is edited in this very diff, so the omission looks deliberate.

**Unresolvable references in shipped comments.** `HouseRulesSheet.tsx`: `/** Variant A's production House rules sheet … */`; `SettingsToggle.tsx`: `/** The 52px house-rules burger from the chosen settings prototype. */`. "Variant A" and "the chosen prototype" exist nowhere in the repo.

### Baseline smells (judgement calls)

- **Middle Man + Speculative Generality** — `rooms.ts` `setSeatCount(code, seatCount) { return this.changeSeatCount(...) }`, no caller. Same in `app.ts`: two routes for one handler (`/rooms/:code/seats/count` and `/rooms/:code/seat-count`), with a comment conceding one is canonical.
- **Duplicated Code + Feature Envy** — `app.ts` repeats `room.seats.find(c => c.claimed && c.token === token)` twice, each reaching into `Room.seats` from outside. Belongs as `RoomStore.findSeatByToken`.
- **Duplicated Code** — `remapEngineState`'s four branches repeat `...state, seats: state.seats.map(map), button: map(state.button)`; `changeSeatCount` repeats the same `{ seatCount, pendingSeatCount: null, applied: true, moves: [] }` literal three times.
- **Data Clumps** — the `{from,to}` move shape is declared five times (`SeatMove`, `SeatMovedMessage`, inline in `table-client/src/api/rooms.ts`, inline in `player-client/src/App.tsx`, `SeatMovePreview`); `SeatCountChange` is declared verbatim in both `rooms.ts` and `table-client/src/api/rooms.ts`. `packages/protocol` is the repo's home for cross-package shapes.
- **Primitive Obsession / inconsistency** — `RoomView.pendingSeatCount?: number` vs `Room.pendingSeatCount: number | null`, forcing `?? null` and a spread-to-omit dance in `toRoomView`.
- **Fit with convention** — `player-client/src/App.tsx` renders the seat-moved notice as a 15-line inline-styled `<div>` with hardcoded `#eef7ef` / `14px`, where the package convention is a small extracted component using `ui-shared` tokens (cf. `InlineError.tsx`).
- **Divergent Change** — `rooms.ts` `dispatch`: the `nextHand` guard changes from `room.engine !== null` to `room.engine?.hand?.status === "complete"`, unrelated to seat count, folded into the single commit.

---

## Spec

### (a) Missing / partial

1. **Table-only route untested.** _"Route: table-only, 400 below the floor, out-of-range rejected as in #74."_ `app.test.ts` covers the floor and range but never asserts table-only — and no gating exists on the route (though `/end` and `/evict` are equally ungated, so the mirroring is consistent).
2. **"out-of-range rejected as in #74" not matched.** `/rooms` distinguishes `invalid-seat-count` from `invalid-request-body`; the new route returns `invalid-seat-count` for _any_ bad body (`app.ts:499-502`).
3. **Growth deal-in untested.** _"Growth applies immediately and deals the new seats in from the next hand."_ `rooms.test.ts` only asserts seats are appended.
4. **Deferral only honours `nextHand`.** ADR-0004: _"applied at the next deal-in recompute at `startHand`/`nextHand`."_ `rooms.ts` applies the pending shrink solely in the `nextHand` branch.

### (b) Scope creep

1. Second route alias `/rooms/:code/seat-count` (`app.ts:527-528`) — the spec asked for one route; one test even calls the alias.
2. Unused `setSeatCount` alias on `RoomStore`.
3. `remapEngineState` of a _completed_ hand (results, winners) plus `broadcastCurrentView` re-sending `view-snapshot` (`app.ts:465-485`) — the spec scoped repack to the deal-in seam.
4. WebSocket auth rework: the pre-handler now resolves the seat by token alone, ignoring the `seat` query param (`app.ts:544-552`). Item 5 asked only that the player "not read it as a glitch".

### (c) Implemented but wrong

1. **`sittingOut` clobbered on the player client.** _"carrying `token`, `sittingOut`, `disconnected` with the player."_ `useWebSocket.ts` `case "seat-moved"` does `setSeat({ seatId: message.to, sittingOut: false })` — a sitting-out player appears sat back in until the next `room-view`.
2. **Repack bounces the player's socket.** `setSeat` → `wsParams` memo (`App.tsx:101-105`) → the `useWebSocket` effect's `params` dep changes → socket torn down and reconnected; the server marks the seat disconnected on close, so a repack flickers `disconnected` for every moved player — the exact "glitch" item 5 forbids.
3. **`nextHand` guard tightened beyond spec** — a `nextHand` dispatched while betting no longer hits the recompute and falls through to the engine.

---

**Summary** — Standards: 2 hard breaches + 7 judgement calls; worst is "House rules" shipped as a domain concept without a `CONTEXT.md` entry in a diff that edits `CONTEXT.md`. Spec: 11 findings; worst is the repack tearing down and reconnecting the player socket, flashing every moved player as `disconnected` — precisely what item 5 was written to prevent.

packages/server/src/app.ts
```diff
@@ -403,12 +461,15 @@ export async function buildApp(
...
+            token === undefined
+              ? undefined
+              : rooms.findSeatByToken(code, token);
           if (
             !seatObj ||
-            !seatObj.claimed ||
+            seatId === undefined ||
             token === undefined ||
-            seatObj.token !== token
+            seatObj.id !== seatId
```
ewanhardingham: **N1 — spec, blocking: a repacked player who is offline can never reconnect.**

This commit added `seatObj.id !== seatId` to the upgrade guard and deleted the reconnect-time `seat-moved` send that `c04c9e0` had in the connect handler (`if (movedFrom !== undefined && identity !== "table")`). `applySeatMoves` only reaches sockets that are *currently open*:

```ts
for (const socket of sockets) {
  if (socketIdentity.get(socket) !== move.from) continue;
```

So a player who is disconnected — or simply reloading — during a repack still holds the stale seat id in `localStorage`. Their reconnect now 403s (asserted at `app.test.ts:565-573`), which fires `onRejected` → `handleRejected` (`player-client/src/App.tsx:65-70`) → `clearSeatToken` + `clearSeat`, dropping them to the seat picker. The server still holds their claim, so nobody can take the seat either.

Issue #77 rule 1: *"The setting never evicts anyone."* This path evicts them and burns the seat. It worked in `c04c9e0`.

The fix for finding 10 (keeping the socket open across a move) was correct — the regression is deleting the reconnect case with it. Suggest: on upgrade, accept a token whose seat has moved, and send `seat-moved` on connect so the client resyncs, rather than rejecting the stale `seat` param.

packages/server/src/rooms.ts
```diff
@@ -220,6 +323,51 @@ export class RoomStore {
...
+      room.pendingSeatCount = null;
+      return appliedSeatCountChange(room);
+    }
+
+    if (seatCount === room.seats.length) {
+      room.pendingSeatCount = null;
+      return appliedSeatCountChange(room);
+    }
+
+    if (room.engine !== null) {
```
ewanhardingham: **N2 — spec drift: deferral widened beyond "mid-hand".**

The queue condition moved from `hand?.status === "betting"` to `room.engine !== null`. `engine` is only ever `null` at room creation, so *every* shrink after the room's first hand is deferred — including when no hand is in progress and the felt is idle.

Issue #77 rule 4: *"A shrink **never applies mid-hand**… Growing is safe immediately."* Acceptance: *"A shrink requested mid-hand applies at the next hand, not immediately."* Between hands is not mid-hand, so the table device now waits for a deal it didn't need to wait for.

ADR-0004 and `CONTEXT.md` were edited in this same commit to say "applies at the next deal-in when a hand is displayed" — i.e. the docs were moved to match the code rather than the issue. The change is defensible (it's what let you drop `remapEngineState`), but it's a spec amendment the issue doesn't authorise. Either narrow the condition back to a live hand, or amend #77 and say so in the ADR's Consequences.

packages/server/src/rooms.ts
```diff
@@ -278,18 +426,35 @@ export class RoomStore {
       return { error: "not-permitted" };
     }
 
+    let seatMoves: readonly SeatMove[] = [];
     if (type === "startHand" && room.engine === null) {
       const dealIn = eligibleSeats(room);
       if (dealIn.length < 2) return { reason: "not-enough-players" };
-      room.engine = createInitialState(dealIn);
+      seatMoves = applyPendingShrink(room).moves;
```
ewanhardingham: **Spec finding 4 — still partial.** `startHand` now calls `applyPendingShrink`, which addresses the ADR-0004 wording (*"applied at the next deal-in recompute at `startHand`/`nextHand`"*). But this branch only runs when `room.engine === null`, and `changeSeatCount` only queues a shrink when `room.engine !== null` — so nothing can ever be pending here in production. The covering test reaches it by setting `room.pendingSeatCount = 2` directly rather than through `changeSeatCount`.

Not wrong, but it's dead defensive code backed by a test that can't fail for the right reason. Worth either deleting or making reachable — resolving N2 would make it reachable.

packages/server/src/app.test.ts
```diff
@@ -342,6 +343,235 @@ describe("seat claim/evict routes", () => {
   });
 });
 
+describe("table-only seat-count settings route", () => {
```
ewanhardingham: **Spec finding 1 — still partial.** The block is now named `"table-only seat-count settings route"`, but no test asserts table-only access, and the route at `app.ts:389` is registered ungated.

Issue #77 Tests: *"Route: table-only, 400 below the floor, out-of-range rejected as in #74."* The floor and range halves are covered; the table-only half is named but not tested.

Fair caveat: `/evict` and `/end` are equally ungated, so the route genuinely does mirror the pattern the issue pointed at. If there's no role gate to test yet, the honest move is to rename the block rather than let it claim a property nothing checks.

packages/server/src/app.ts
```diff
@@ -330,6 +353,41 @@ export async function buildApp(
...
   });
 
+  /** Table-device house rules: change the room's seat count (issue #77). */
+  const changeSeatCount = (
+    request: FastifyRequest<RoomCodeRoute>,
+    reply: FastifyReply,
+  ) => {
+    const body = ChangeSeatCountRequestSchema.safeParse(request.body);
+    if (!body.success) {
+      const blamesSeatCount = body.error.issues.some(
```
ewanhardingham: **Standards, judgement call — Duplicated Code.** This zod-blame block is verbatim the one in the `POST /rooms` handler:

```ts
const blamesSeatCount = body.error.issues.some(
  (issue) => issue.path[0] === "seatCount",
);
```

Two copies of the same "which field did zod object to" decision. Extract one `seatCountBodyError(result)` and call it from both — the next route that takes a seat count will otherwise make it three.

packages/server/src/rooms.ts
```diff
@@ -278,18 +426,35 @@ export class RoomStore {
...
+      }
+
       const dealIn = eligibleSeats(room);
-      if (dealIn.length < 2) return { reason: "not-enough-players" };
-      room.engine = {
-        ...room.engine,
-        seats: dealIn,
-        button: resolveButtonFor(room.engine.button, dealIn),
-      };
+      if (room.engine.hand?.status === "complete") {
```
ewanhardingham: **Standards, judgement call — dead temp + Duplicated Code.** `const dealIn = eligibleSeats(room)` is computed just above, then this branch ignores it and recomputes `const nextDealIn = eligibleSeats(room)` after the repack (correctly — the repack changes the answer). The `else` branch uses `dealIn`. Both then build the same literal shape:

```ts
room.engine = { ...room.engine, seats: <dealIn>, button: resolveButtonFor(<prev>, <dealIn>) };
```

Hoist the common assignment and vary only the seats/previous-button pair, and move the `dealIn` binding into the branch that actually uses it. The same pattern repeats in `startHand`, where `dealIn` serves only the `< 2` check and `createInitialState(eligibleSeats(room))` re-queries.

packages/server/src/rooms.ts
```diff
@@ -47,13 +49,22 @@ export interface Room {
...
+  /** A live-hand shrink waits here until the next deal-in recompute. */
+  pendingSeatCount: number | null;
 }
 
 export type ClaimSeatError =
   "room-not-found" | "seat-not-found" | "seat-already-claimed";
 
 export type ClaimSeatResult = { seat: Seat } | { error: ClaimSeatError };
 
+export type ChangeSeatCountError =
```
ewanhardingham: **Standards, judgement call — Primitive Obsession / union drift.** `ChangeSeatCountError` lists three members, but the route can also send `"invalid-request-body"` as a bare string literal (`app.ts:367`). The set of errors a client must handle isn't typed anywhere, so the two can drift silently.

Either widen this union to cover the wire errors and have the handler return a value of it, or give the transport its own error union in `packages/protocol` alongside `SeatCountChange`.

packages/player-client/src/ws/useWebSocket.ts
```diff
@@ -61,23 +64,42 @@ export function useWebSocket(
...
   const commandRejected = usePlayerStore((state) => state.commandRejected);
   const socketRef = useRef<WebSocket | null>(null);
   const optionsRef = useRef(options);
+  const currentSeatIdRef = useRef<number | null>(params?.seatId ?? null);
   optionsRef.current = options;
+  currentSeatIdRef.current = params?.seatId ?? null;
 
+  const roomCode = params?.roomCode ?? null;
+  const token = params?.token ?? null;
+  const hasParams =
```
ewanhardingham: **Standards, judgement call — duplicated state + a redundant flag.** The current seat is now tracked three ways: `currentSeatIdRef` (written during render, line 77), an effect-local `let currentSeatId`, and the store via `moveSeat`. That's the right shape for keeping the socket alive across a move — the concern is that three writers must stay in step by hand.

`hasParams` is also redundant: it re-derives a null check that `params` already encodes, and the `initialSeatId === null` guard a few lines below repeats the seat half of it. Deriving `roomCode`/`token` from a single narrowed `params` would collapse both.

packages/player-client/src/SeatMovedNotice.tsx
```diff
@@ -0,0 +1,26 @@
...
+export interface SeatMovedNoticeProps {
+  readonly message: string;
+}
+
+/** The inline notice shown when a table repack changes a player's position. */
+export function SeatMovedNotice({ message }: SeatMovedNoticeProps) {
+  return (
+    <div
+      data-testid="seat-moved-notice"
+      style={{
```
ewanhardingham: **Standards, judgement call — convention nit.** Good extraction, and the `ui-shared` tokens are exactly what the first pass asked for. Two small departures from `InlineError.tsx`, the component this mirrors: that file hoists a module-level `const style: CSSProperties` rather than rebuilding the object each render, and `margin: "10px 18px 0"` / `padding: "10px 14px"` are magic values where the rest of the styling is tokenised.

ewanhardingham (commented):
## Code review — second pass (`7dc0e62`)

Re-reviewed `origin/main...issue-77-seat-count` on both axes. Findings are inline; this is the roll-up.

**Standards — 9/9 first-pass findings resolved.** `CONTEXT.md` gains a **House rules** entry; the prototype-era comments are rewritten; the `setSeatCount` and `/seat-count` aliases are gone; `findSeatByToken` lands on `RoomStore`; `remapEngineState` is deleted; `SeatMove`/`SeatCountChange` moved to `packages/protocol`; `pendingSeatCount` nullability unified; `SeatMovedNotice` extracted onto `ui-shared` tokens; the `nextHand` guard restored. No hard breaches remain — the five inline standards notes are all judgement calls.

**Spec — 8 resolved, 3 partial, 2 new.**

| First-pass finding | Status |
|---|---|
| 1. Table-only route untested | **partial** |
| 2. out-of-range vs bad body | resolved |
| 3. Growth deal-in untested | resolved |
| 4. Deferral only honours `nextHand` | **partial** |
| 5-7. Route/method aliases, completed-hand remap | resolved |
| 8. WS auth by token alone | **partial** — see N1 |
| 9. `sittingOut` clobbered | resolved |
| 10. Repack bounces the socket | resolved |
| 11. `nextHand` guard tightened | resolved |

Server and player-client suites pass (190 tests).

**The one blocker is N1** (`app.ts`, inline): fixing finding 10 by keeping the socket open was right, but deleting the reconnect-time `seat-moved` alongside it means a player who is *offline* during a repack can never get back in — and the client clears their token on the 403, so the seat becomes unreachable by anyone. That is the outcome rule 1 ("the setting never evicts anyone") exists to prevent. N2 is a smaller spec-drift call for you to make deliberately rather than by omission.

Issue #86, opened 2026-07-31
Does the live server write the hand log at all?


ewanhardingham:
## Question

Map [#79](https://github.com/ewanhardingham/table-top-poker/issues/79) says Phase 1 "writes the log but never reads it back — the raw event log is server-side only". That premise is wrong, and the whole map rests on it.

The log writer, `HandLog` (`packages/harness/src/persistence.ts`), lives in **`harness`** and is used only by the harness CLI. The server has no dependency on `harness`, no `node:fs` write of any kind, and no `logDir` concept:

```
$ grep -rn "writeFile\|appendFile\|createWriteStream\|mkdir" packages/server/src/   # nothing
$ grep -c harness packages/server/package.json                                      # 0
```

So a live room **persists nothing**. Phase 2 cannot serve a session's recorded hands back to the table until the server starts writing them, and no ticket on this map currently covers that.

- Does the server reuse `HandLog`, or does the writer move? `harness` depends on `engine`; a `server → harness` dependency needs checking against the structurally-enforced dependency direction ([issue #3](https://github.com/ewanhardingham/table-top-poker/issues/3)), and dragging the CLI's stdin/stdout harness into the server to get at its persistence may be the wrong seam.
- What is a room's game id and log directory? The on-disk shape is one directory per game ([Phase 1 spec §5](https://github.com/ewanhardingham/table-top-poker/issues/1)) with a `game.jsonl` seats manifest; a room has a rotating join code, not a stable id.
- The manifest is written once at construction from a fixed seat list, but seats are claimed, evicted and sat out over a session's life, and seat count is now chosen at room creation. Does the manifest still describe a live room accurately?
- `HandLog` writes every record with a synchronous `appendFileSync`. In the harness that is a CLI on one hand at a time; in the server it is on the socket hot path with the action clock running. Acceptable, or does it need moving off the request path — and if so, what happens to the durability property (`packages/harness/src/persistence.durability.test.ts`) that a killed process loses at most one record?
- Is the log written for every room unconditionally, or opt-in? A LAN box hosting casual games accumulating every hand forever is a different product decision from a debugging aid.

Blocks the wire contract ticket — there is nothing to serve until this is answered.

### Downstream replay constraint

[...]

Issue #80, opened 2026-07-31
Replay capability: API, package placement and version-tag handling


ewanhardingham:
## Question

The shared middle both surfaces depend on. What is the replay capability, where does it live, and how does it behave against a log it was not built for?

- What does replay expose at an event ordinal — the full `state`, a projected `view`, both? Does it return one position, or fold over a range?
- The engine is pure (no I/O, no clock, no ambient randomness), so file reading cannot live there. Does the pure part sit in `engine` with I/O in `harness`/`server`, or does a new package own it? Dependency direction is enforced structurally ([issue #3](https://github.com/ewanhardingham/table-top-poker/issues/3)), and `engine` is strict-TDD.
- Replay is bit-identical only within a matching engine build, and every log record carries a version tag ([issue #10](https://github.com/ewanhardingham/table-top-poker/issues/10)). What happens when the tag does not match the running build? Within a live session this cannot arise — a room is one process — so this is really a dev-stepper concern, but the answer belongs with the capability rather than in the CLI ticket.
- Does the capability replay the *command* log through `decide`, or fold the *event* log through `apply`? These are different trust models: re-deriving would catch a corrupt event log, folding trusts it and is cheaper.
- Append-as-you-go means the last line of a hand's log may be torn. Does the capability tolerate a truncated trailing record, or refuse the file?

Standing constraints from the map: position is an event ordinal; replay re-projects `view(state, seatId)` and cannot leak; the visibility split is by surface, never a flag on the data.

ewanhardingham:
Constraint from the hand picker ticket ([#81](https://github.com/ewanhardingham/table-top-poker/issues/81)), which the capability has to satisfy:

**A terminal view is not enough — the picker needs a fold over the events.** `FoldedOutView` (`packages/engine/src/view.ts:21`) carries only `button` and `winner`. It has no board, so the board of a hand that ended by fold-out is not recoverable from the hand's final `view` at all, even though `BoardDealt` was public and the table watched those cards land. The chosen picker renders that board as its primary identifying feature, and derives a betting-shape phrase ("raise war — 4 raises") from the public `ActionTaken` events.

So whatever the capability exposes must let a caller reach *public events across a whole hand*, not just the state at one ordinal. Worth deciding here whether that is a separate summary-shaped entry point or falls out of "fold over a range".

ewanhardingham:
## Resolution

### Capability and ownership

[...]

Issue #79, opened 2026-07-31
Map: session replay — a build-ready Phase 2 specification


ewanhardingham:
> **Destination reached.** Every ticket on this map is closed and the way to a build-ready Phase 2 is clear. The specification is [**Phase 2 build-ready specification** (#129)](https://github.com/ewanhardingham/table-top-poker/issues/129). Nothing below under **Not yet specified** is unfinished work — it is fog beyond this destination, and each item would open its own map.

## Destination

A **build-ready Phase 2 specification**: session replay for recorded hands, built as one shared replay capability with two surfaces — a table-device review of the current session's hands, and a non-interactive dev stepper over any log directory on disk.

Phase 2 is [the second phase on the roadmap](https://github.com/ewanhardingham/table-top-poker/issues/5), promoted ahead of tactile interactions and chips because its debugging value (diagnosing engine behaviour while building every later phase) and its player-facing value (reviewing a hand before the next one starts) both outrank them, and it depends on neither.

## Notes

**Where this starts from.** [Phase 1](https://github.com/ewanhardingham/table-top-poker/issues/1) does not persist a live room at all — corrected while resolving the hand picker ticket. The log writer `HandLog` lives in `harness` and is used only by the harness CLI; the server has no `harness` dependency and no filesystem write of any kind. So Phase 2 opens a read path **and** a write path, and the write path is charted as [Does the live server write the hand log at all?](https://github.com/ewanhardingham/table-top-poker/issues/86).

**Mechanical replay already works.** `packages/harness/src/replay.test.ts` proves that re-piping a persisted command log through the harness reproduces the persisted event log exactly. Phase 2 is not building replay; it is building the tooling on top of it.

**On-disk shape** is settled by [Does the live server write the hand log at all?](https://github.com/ewanhardingham/table-top-poker/issues/86): a Room directory contains immutable Room metadata plus an independently replayable context/Command/Event triplet per Hand. Writes are append-as-you-go; abrupt death may tear only the final record.

**Decisions locked while charting** — binding on every ticket:

- **Replay obeys live visibility.** Replay re-projects `view(state, seatId)`; muck stays mucked, folded hole cards are never shown to anyone, showdown reveals exactly what showdown revealed. Phase 1's [fast-check secrecy property test](https://github.com/ewanhardingham/table-top-poker/issues/11) carries into Phase 2 unchanged — there is still no view type structurally capable of holding another seat's cards.
[...]

Issue #35, opened 2026-07-25
17 — Acceptance run


ewanhardingham:
## Source

Phase 1 spec #130 §11 (Acceptance).

## What to build

The full Phase 1 acceptance checklist, run in one sitting, on real devices at a real table — confirming every prior ticket's piece actually holds together end-to-end, not just in isolation.

## Acceptance criteria

- [x] The table device creates a room; its QR code is scanned successfully by a phone on the main SSID (not an isolated guest network).
- [x] At least 3 phones claim distinct seats via the picker; the table device shows all claimed seats before the first hand starts.
- [x] A hand is started with a live seed; each phone shows only its own hole cards — never another seat's, at any point before showdown.
-  [x] A full betting round completes on every street (preflop through river) with fold/check/call/raise all exercised at least once, and street closure correctly waits for every live player to act since the last bet/raise.
- [x] At least one hand ends by fold-out before river — no showdown, no reveal, and the table device moves straight to `HAND_COMPLETE`.
- [x] At least one hand reaches showdown — the table device displays the winning hand(s), each live seat's best five cards and rank, and a split is correctly reported if hands tie.
- [x] Pressing "Next hand" on the table device rotates the button and starts a new hand with a fresh seed.
- [x] One phone is locked mid-hand (screen off, socket drops) when it is not that player's turn, and the player is not auto-folded — the fold only fires later, when the action clock reaches them without a response.
- [ ] The completed hands are present on disk afterward as JSON Lines files (seed + command list, and event stream), and replaying one through the harness reproduces the same event stream.

## Blocked by

01–16 (all prior Phase 1 tickets).

PR #93, opened 2026-07-31
docs: assemble the Phase 2 specification

### LINKED ISSUE #85: Assemble the Phase 2 specification
## Question

**This map's destination.** Gather every decision above into `docs/phase-2-spec.md` — a document a build session can work from without re-deriving anything, in the shape of the Phase 1 spec (#130).

Expected contents:

- Scope boundary, including what is explicitly out (crash recovery, full-reveal modes, phone-side replay).
- The replay capability: API, placement, trust model, version-tag behaviour.
- The visibility model in replay, and why Phase 1's secrecy property test still holds unchanged.
- The wire contract for serving a hand back.
- The table-device review: picker, playback, when it is reachable, how it is dismissed.
- The dev stepper CLI.
- An acceptance checklist — the demonstrable thing that proves Phase 2 is done.

Assembly is also the moment contradictions surface. Phase 1's assembly found two stale statements in `CONTEXT.md`; expect the same here, and fix rather than document them.

Blocked on every other ticket.

ewanhardingham:
Resolves the final ticket on [wayfinder map #79](https://github.com/ewanhardingham/table-top-poker/issues/79) — [Assemble the Phase 2 specification](https://github.com/ewanhardingham/table-top-poker/issues/85), which was blocked on all six other tickets and is now unblocked.

## What this adds

`docs/phase-2-spec.md`, in the shape of `docs/phase-1-spec.md` — one document a build session can work from without re-deriving anything:

1. **Scope** — destination, the "Phase 1 never persisted a live Room" starting position, and the out-of-scope boundary (crash recovery, full-reveal modes, phone-side replay).
2. **Standing constraints** — replay obeys live visibility, position is an event ordinal, the visibility split is by surface not by a flag, rejections are excluded structurally.
3. **Room recording** — the `@table-top-poker/recording` package, Room ID layout, operation-oriented append, the recording-paused state.
4. **The replay capability** — engine-owned, per-Hand flipbook, validation, context validity, version rules, the two incomplete-Hand cases (a torn final record, and an orphaned trailing Command), and the terminal-view finding that forces event+state per position.
5. **Wire contract** — replay over the existing room socket, whole hand in one response, proactively-pushed hand summaries.
6. **The table review** — reachability and force-dismiss, the filmstrip picker, the live-ticking relative clock, the chaptered scrub.
7. **The dev stepper CLI** — `harness replay`, JSONL out, exit 0/1/2.
8. **Corrections** to existing docs.
9. **Acceptance** — a demonstrable checklist across write path, replay, table review, and the standing secrecy guarantee.

## Contradictions found during assembly, fixed rather than documented

- **`CONTEXT.md` was missing every Phase 2 term.** The resolutions of #80 and #86 both state that `CONTEXT.md` "now defines" Replay, Replay position, Rejection, Room ID, Room recording, Hand context and Hand recording — no such edit ever landed on `main`. Added.
- **`docs/phase-1-spec.md` describes a server-side log the server never wrote.** §5 specifies a write path that shipped only in `packages/harness`; §1 and §4 repeat the claim. Corrected in place, pointing at Phase 2 §3.
- **Two version numbers, deliberately separated.** #86's "recording format version 2" and #80's "must carry the running engine's `ENGINE_LOG_VERSION`" are *different* tags with different meanings, and neither ticket noticed the other. `ENGINE_LOG_VERSION` **stays `1`** — Phase 2 changes no `HandEvent` shape, no `Command` shape and no shuffle, so a `v: 1` record replays bit-identically, which is exactly what that constant is documented to mean. The recording **layout** version lives on `room.json` alone. (Assembly first collapsed these into one tag; grilling reversed it — see the review comment below.)

Docs only — no code changes. `prettier --check` passes on all three files.

Closes #85

https://claude.ai/code/session_01CfXs8UqpShQ3qcQWBLdK6x

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -145,8 +145,10 @@ _Avoid_: Using the join code as a durable identifier.
 **Room recording**:
 The durable, unredacted, local record of a Room — an immutable `room.json`
 manifest plus one Hand recording per Hand, under the Room ID's directory.
-Written by the server for every Room, always, from creation until the Room
-ends. Owned by the `recording` package; the engine holds no I/O.
+Written by the server for every Room, from creation until the Room ends. It
+cannot be disabled by configuration and never stops silently; it stops only
+where the table is told a write failed and deliberately chooses to carry on
+without it. Owned by the `recording` package; the engine holds no I/O.
 _Avoid_: Log — a Room recording is the whole directory, not one file.
 
 **Hand context**:
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -128,6 +128,52 @@ Phase 1 tracks no value, no state; the engine never references a Pot.
 _Avoid_: Treating Pot, Stack, Chips, or Side pot as engine concepts —
 they're Phase 4 scope, and conditional even then.
 
+**Rejection**:
+The typed value the engine returns when a Command is not valid —
+`{ type: 'Rejection', reason, command }`. Never thrown, never itself an
+Event, never processed by `apply`. Delivered only to the Device that sent
+the Command, and kept in the Hand recording for audit; it is absent from
+every player-facing shape, including Replay.
+_Avoid_: Calling a Rejection an error or an event.
+
+**Room ID**:
+The opaque UUID a Room is given at creation, naming its Room recording
+directory on disk. Distinct from the four-character join code, which is a
+live, human-typed handle with no durable meaning.
+_Avoid_: Using the join code as a durable identifier.
+
+**Room recording**:
+The durable, unredacted, local record of a Room — an immutable `room.json`
+manifest plus one Hand recording per Hand, under the Room ID's directory.
+Written by the server for every Room, always, from creation until the Room
+ends. Owned by the `recording` package; the engine holds no I/O.
+_Avoid_: Log — a Room recording is the whole directory, not one file.
+
+**Hand context**:
+The immutable document a Hand recording opens with: the participating
+Seats, the starting Button, the Hand ordinal, and `startedAt`. Contains no
+cards and no state snapshot — it is the bootstrap Replay needs, not a
+saved position.
+
+**Hand recording**:
+One Hand's independently replayable triplet within a Room recording — its
+Hand context, its exact ordered Commands, and the resulting Events and
+Rejections. Replaying it never requires any other Hand.
+
+**Replay**:
+Rebuilding a Hand from its Hand recording by re-running its Commands
+through the engine and validating the generated Events against the
+persisted ones. A pure engine capability. Replay obeys live visibility
+exactly: it re-projects `view`, so a muck stays mucked and a folded Seat's
+cards are shown to no one.
+_Avoid_: Treating Replay as a stored playback or a recording of views.
+
+**Replay position**:
+A point within a replayed Hand, addressed as an Event ordinal — position
+*n* is the state after applying *n* Events, and carries that *n*th Event
+alongside it. Position 0 is the starting state with no Event. A Rejection
+occurs *at* a position without advancing it.
+
 ## Hand lifecycle
 
 ```
--- a/docs/phase-1-spec.md
+++ b/docs/phase-1-spec.md
@@ -245,12 +245,15 @@ network round trip.
 - **Retention**: keep everything, forever. No rotation or size cap for
   Phase 1.
 
-> **Correction — this section specifies a write path Phase 1 never shipped
-> in production.** As built, the only writer is `HandLog`
+> **Correction — this section specifies a write path Phase 1 did not ship in
+> production.** As merged on `main`, the only writer is `HandLog`
 > (`packages/harness/src/persistence.ts`), used by the harness CLI alone.
 > `packages/server` has no `harness` dependency and performs no filesystem
 > write of any kind, so a Room played through the server left nothing on
-> disk. Discovered while charting
+> disk — which is why §11's last acceptance box is still unchecked
+> ([#35](https://github.com/ewanhardingham/table-top-poker/issues/35)).
+> Branch `fix/issue-35-hand-persistence` is closing that gap. Discovered
+> while charting
 > [map #79](https://github.com/ewanhardingham/table-top-poker/issues/79).
 > The durable server-side write path is delivered by Phase 2 —
 > [`docs/phase-2-spec.md`](phase-2-spec.md) §3 — which also supersedes the
--- a/docs/phase-1-spec.md
+++ b/docs/phase-1-spec.md
@@ -34,7 +34,8 @@ on each):
   Phase 1's UI is tap-a-button; the frontend choice must not preclude Phase 3,
   but building it is not this phase's job.
 - Replay and audit tooling (browsing/stepping through recorded hands) —
-  Phase 2. Phase 1 writes the log; it does not read it back.
+  Phase 2. Phase 1 writes the log (in the harness only — see the correction
+  in §5); it does not read it back.
 - Public internet deployment, accounts, authentication — later. Phase 1 is
   LAN-only, no auth, bearer-secret trust model throughout.
 - Poker variants other than Texas hold'em, and multiple simultaneous tables.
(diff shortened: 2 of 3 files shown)


ewanhardingham:
## Reflog check: the terms were never written, but something else turned up

Went looking for whether the `CONTEXT.md` terms had been written and lost rather than never written. **They were never written** — searched every reachable commit on every ref, every dangling commit, and every dangling blob in the object store for `Replay position` / `Room recording` / `Hand recording`; the only hit is my own commit on this branch. `git log --all -- CONTEXT.md` shows 18 commits touching the file and none of them is a glossary addition for these terms. So §8's first correction stands as written.

## What the reflog did surface

`fix/issue-35-hand-persistence` (`ff9ae12`, pushed to `origin` last night, **no PR, not merged**) — which closes the last unchecked box of Phase 1's acceptance run ([#35](https://github.com/ewanhardingham/table-top-poker/issues/35)) by lifting `HandLog` out of `harness` into a new `@table-top-poker/persistence` package and calling it from the server.

That doesn't change any decision on [map #79](https://github.com/ewanhardingham/table-top-poker/issues/79), but it makes the spec's starting position stale in a way that would have cost a build session real time: §3 said "create a new `@table-top-poker/recording` package", which would have produced a *third* package alongside `harness`'s old one and this new one.

Pushed a second commit reconciling it:

- **§1** now names the branch as in-flight rather than asserting the server writes nothing.
- **§3** carries a reconciliation table — **adopt and evolve `@table-top-poker/persistence`, don't add a third package** (the name is a free call; nothing rides on it) — plus what it does not yet satisfy:

| §3 requires | `ff9ae12` as built |
| --- | --- |
| Durable **Room ID** directory | keyed by the join code, which is re-rolled on collision |
| Recording is a **Room invariant** | opt-in; `handLogDir === undefined` silently disables it |
| Transactional `room.json` | no manifest at all |
| **Hand context** with `startedAt` | absent — so §6's picker clock has no source |
| Stage → await append → commit | fire-and-forget after dispatch |
| **recording-paused** on failure | no failure path |
| Ordered async queue, confirmed offsets | synchronous `appendFileSync` |
| `v: 2` | `v: 1` |

The gap is narrow in kind: that branch is the *file-writing* half, Phase 2 adds the *durability and identity* half around it. Nothing in it needs reverting.

- **§8** and the `docs/phase-1-spec.md` §5 correction now say "as merged on `main`" rather than implying the gap is unaddressed, and point at #35.

ewanhardingham:
## Four clarifications the reconciliation exposed

Re-read the spec asking what the persistence finding had made ambiguous. Four gaps, each one a place where two closed tickets met and neither had reason to notice the other. None changes a decision on [map #79](https://github.com/ewanhardingham/table-top-poker/issues/79); all four are pushed.

**1. Harness-written recordings had no Room ID story.** §7's stepper reads a Room recording directory, and the Room recordings a developer has locally are mostly harness-produced — but a harness run has no Room, no join code and no Room ID, and §3 only said "the harness retains `--log-dir`". As written, `harness replay` could not read what `harness --log-dir` had just written. Now specified: `--game-id` (already defaulting to a timestamp) becomes the Room ID, `room.json`'s `code` is `null`, and the harness emits context sidecars like the server does.

**2. `recording-paused` would have landed in the engine.** §3 called it "a reason code", and Phase 1 fixed `Rejection.reason` as a *closed, engine-defined* set — so the obvious reading puts it in `packages/engine/src/types.ts` and teaches the pure engine about filesystem failure. It belongs in `ServerRejectionReason` (`packages/protocol/src/hand.ts`), next to `room-not-found`. A failed write is not a rule violation and `decide` never sees one. Said explicitly now.

**3. A reloaded table device would have shown an empty picker.** §5 had summaries pushed "right after each hand completes", and Phase 1's catch-up is deliberately one fresh view snapshot carrying no summaries — so a table that reloads mid-session loses every summary it had. Surviving a client reload was one of the map's stated reasons for serving a durable log instead of buffering client-side, so this would have quietly broken the thing the decision was made for. The server now sends the accumulated list on connect; that's what the `list-hands` shape is for.

**4. The one-version-tag reading is mine, not the map's.** #86 and #80 each specified a version tag without noticing the other. Collapsing them is the cheaper reading and I still recommend it, but it is an assembly reconciliation and the spec now says so — including its wart (a recording-format change now forces an engine log version bump) and the two things `1 → 2` invalidates: the `persistence.test.ts` fixtures, and any logs the in-flight branch has already written.

Also folded in: `HAND_LOG_DIR` vs `RECORDINGS_DIR` is a **deployment** change — `/etc/poker/poker.env` on the Pi has to move in the same step or recording silently falls back to the default directory. Keeping the existing name is fine; the opt-in behaviour is not.

Two new acceptance boxes cover the harness round-trip and the table-reload case.

ewanhardingham:
Grilled this before merge; fourteen decisions came out of it and are now folded into the spec (`50b0136`). Summary of what changed, and two issues raised along the way.

**Note:** the PR description's third bullet — *"One version number, not two"* — is now **inverted** by this revision. Worth editing before merge.

### §3 — the write path

- **recording-paused gains a third exit: Continue without recording.** The failure this state most likely meets is a worn SD card flipping the filesystem read-only, where every Retry fails identically and the only working control was End session. The invariant is now precisely **no *silently* unrecorded play**.
- **The commit seam is a transaction handle** — `RoomStore` produces it, `app.ts` sequences `append` → `commit`/`discard`. Staging must cover **seat repacks and `pendingSeatCount`**, not just the engine transition; `dispatch` mutates all three before `decide` runs today.
- **A queued clock-fold carries a room revision**, not just its expected Actor. Heads-up, the non-button seat acts *last* preflop and *first* on the flop (`table.ts:49-63`), so actor-matching alone can fold a hand its owner just voluntarily checked. Clock arms after commit; Retry's fresh interval is stated as a player-facing rule.
- **The version tags split.** `ENGINE_LOG_VERSION` **stays `1`** — Phase 2 changes no event shape, command shape or shuffle, so a `v: 1` record replays bit-identically, which is exactly what the constant is documented to mean. The layout version lives on `room.json` alone. The old collapse also justified itself with "the server produced no version-1 recordings", which died when #94 merged.
- **The package and its vocabulary are renamed.** This PR adds a `CONTEXT.md` that says *"Avoid: Log"* and defines Room recording / Hand recording / Room ID — while the package it governs exports `HandLog`, `gameId`, `game.jsonl`, `HAND_LOG_DIR`. Deferring the rename would ship a glossary contradicted by every identifier under it. Filesystem operations become injectable, following `ActionClock`'s precedent, so the recovery paths are testable.

### §4 / §7 — replay and the stepper

- **An orphaned trailing Command is incomplete, not corrupt** (exit `2`, prefix shown). It is what a `SIGKILL` leaves *and* what a read-only disk leaves — and §3's repairs are themselves writes, so they cannot run in that state.
- **Context validity is defined** (`button ∈ seats`, 2–8 seats) — §7 promised an "invalid context" failure with nothing specifying one. Replay constructs its starting state internally rather than adding an engine export.
- **The positional argument accepts a path, a join code, or `latest`.** Codes are recycled (`generateRoomCode` checks live rooms only), so the resolution logs its choice to stderr and stdout stays diffable.

### §5 / §6 — wire and table

- **`summarise(events, context)` is one pure shared function**, removing silent drift between the live and replay derivations of the same facts.
- **`bettingShape` is a structured descriptor**, wording stays client-side. §7 exists so an agent can diff the output; `"raise war — 4 raises"` would force a JSONL consumer to regex English back into a number.
- **Hand length is unbounded** — `raise` is always legal and uncapped, and Phase 1 has no chips. Accepted rather than capped; real byte figures now stated (a `TableView` is ~250–350 bytes, a hand under 20 KB), and chapters rather than ticks are the navigation contract.

### Withdrawn during the grilling

Disk exhaustion (~10 KB/hand — not a real risk) and summary durability across a server restart (`RoomStore` holds `#rooms` in a plain `Map` with no rehydration, so a restart destroys the Room outright — that's crash recovery, out of scope).

### Raised separately

- #95 — a **live bug on `main`**: a `nextHand` arriving mid-hand rewrites `room.engine.seats` and `button`, *then* gets rejected `stale-next-hand`, leaving a seated player silently cut off from every broadcast and auto-folded. No test for `stale-next-hand` exists anywhere in `packages/server`.
- #96 — three Phase 1 rail/pill fixes stranded on the prototype branches, which should land before Phase 2 starts since it adds **Review hands** to that same rail.

PR #100, opened 2026-08-01
fix(server): auto-fold evicted current actor

### LINKED ISSUE #90: Evicting the player to act leaves the game hung on that seat
## Bug

When the table device evicts a seat via \`rooms.ts\`'s \`evictSeat\` (ADR-0003's manual evict action) while that seat is the player currently "to act", the game hangs: the seat is freed back into the join picker, but nothing advances the engine's turn to the next seat. The hand is left waiting on a seat that no longer has an occupant.

## Where

\`evictSeat\` (packages/server/src/rooms.ts:216) just calls \`freeSeat(seat)\` — it never touches engine turn state:

\`\`\`ts
evictSeat(code: string, seatId: SeatId): void {
  const room = this.#rooms.get(code);
  const seat = room?.seats[seatId];
  if (!seat) return;
  freeSeat(seat);
}
\`\`\`

Compare to the voluntary sit-out path, which is documented as deliberately *not* reaching the engine (excluded only at the next deal-in recompute) — eviction has no such excuse, since it can happen mid-hand while a seat is actively "to act".

## Repro

1. Start a hand with seat evicted.
2. Wait until it's seat X's turn to act.
3. Evict seat X from the table device.
4. Observe: seat is freed/unclaimed, but the game state still shows seat X as "to act" — no fold/timeout/skip occurs, and the hand cannot progress.

## Expected

Evicting the player currently to act should force the engine to resolve that turn (e.g. auto-fold or equivalent) and advance to the next seat, same as any other mid-hand player-can't-act path.

## Suggested area

\`evictSeat\` in packages/server/src/rooms.ts needs to check whether \`seatId\` is the engine's current actor and, if so, drive the engine forward before/while freeing the seat.

ewanhardingham:
Closes #90

## Summary

- auto-fold the current actor before freeing an evicted seat
- broadcast the synthesized fold and advance the action clock
- preserve the current actor's timeout when another seat is evicted
- add RoomStore, WebSocket, and action-clock regression coverage

## Verification

- `npm run build`
- `npm run lint`
- `npm test` — 66 files, 393 tests

### DIFF
--- a/packages/server/src/app.test.ts
+++ b/packages/server/src/app.test.ts
@@ -1525,12 +1525,12 @@ describe("action clock", () => {
     const actor = rooms.currentActor(room.code);
     if (actor === undefined) throw new Error("expected a current actor");
 
-    await settle(120);
+    await settle(150);
     const evicted = await app.inject({
       method: "POST",
       url: `/rooms/${room.code}/seats/${String((actor + 1) % 3)}/evict`,
     });
-    await settle(130);
+    await settle(120);
 
     expect(evicted.statusCode).toBe(204);
     expect(actionsSeen(table.messages)).toContainEqual(
--- a/packages/server/src/app.test.ts
+++ b/packages/server/src/app.test.ts
@@ -1173,6 +1173,47 @@ describe("hand command dispatch over WebSocket", () => {
     expect(lastView.view.seats[1]).toMatchObject({ claimed: false });
   });
 
+  it("auto-folds an evicted current actor and broadcasts the turn advance", async () => {
+    const room = rooms.create();
+    const table = connect(`room=${room.code}&role=table`);
+    await opened(table.socket);
+    const seatSockets = [
+      await claimAndConnect(room.code, 0),
+      await claimAndConnect(room.code, 1),
+      await claimAndConnect(room.code, 2),
+    ];
+    await settle();
+
+    table.socket.send(JSON.stringify({ type: "startHand" }));
+    await settle();
+    const actor = rooms.currentActor(room.code);
+    if (actor === undefined) throw new Error("expected a current actor");
+
+    const closed = new Promise<void>((resolve) => {
+      seatSockets[actor]?.socket.once("close", () => {
+        resolve();
+      });
+    });
+    const evicted = await app.inject({
+      method: "POST",
+      url: `/rooms/${room.code}/seats/${String(actor)}/evict`,
+    });
+    await closed;
+    await settle();
+
+    expect(evicted.statusCode).toBe(204);
+    expect(rooms.get(room.code)?.seats[actor]).toMatchObject({
+      claimed: false,
+    });
+    expect(rooms.currentActor(room.code)).not.toBe(actor);
+    expect(table.messages).toContainEqual(
+      expect.objectContaining({
+        type: "hand-update",
+        event: { type: "ActionTaken", seatId: actor, action: "fold" },
+      }),
+    );
+  });
+
   it("rejects an out-of-turn action, visible only to the sender", async () => {
     const room = rooms.create();
     const table = connect(`room=${room.code}&role=table`);
@@ -1470,6 +1511,31 @@ describe("action clock", () => {
         action: "fold",
       }),
     ]);
+  });
+
+  it("does not reset the current actor's clock for a different seat's eviction", async () => {
+    const room = rooms.create();
+    const table = connect(`room=${room.code}&role=table`);
+    await opened(table.socket);
+    for (const seatId of [0, 1, 2]) rooms.claimSeat(room.code, seatId);
+    await settle();
+
+    table.socket.send(JSON.stringify({ type: "startHand" }));
+    await settle();
+    const actor = rooms.currentActor(room.code);
+    if (actor === undefined) throw new Error("expected a current actor");
+
+    await settle(120);
+    const evicted = await app.inject({
+      method: "POST",
+      url: `/rooms/${room.code}/seats/${String((actor + 1) % 3)}/evict`,
+    });
+    await settle(130);
+
+    expect(evicted.statusCode).toBe(204);
+    expect(actionsSeen(table.messages)).toContainEqual(
+      expect.objectContaining({ seatId: actor, action: "fold" }),
+    );
   });
 
   it("does not auto-fold a seat that acts before the clock elapses", async () => {
--- a/packages/server/src/app.ts
+++ b/packages/server/src/app.ts
@@ -406,19 +406,21 @@ export async function buildApp(
       // `actor` was read as the live current actor at schedule time, and
       // any real action in between would have rescheduled (and thus
       // replaced) this very timer — so `dispatch` rejecting the
-      // synthesized fold isn't expected to happen. `for` runs zero times
-      // if it somehow does, and the clock still re-arms below.
+      // synthesized fold isn't expected to happen. If it somehow does, the
+      // clock still re-arms below.
       if ("steps" in result) {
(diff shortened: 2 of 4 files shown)


packages/server/src/rooms.ts
```diff
@@ -367,14 +372,22 @@ export class RoomStore {
...
+  evictSeat(code: string, seatId: SeatId): EvictSeatResult {
     const room = this.#rooms.get(code);
     const seat = room?.seats[seatId];
-    if (!seat) return;
+    if (!seat) return {};
+
+    let fold: DispatchSuccess | undefined;
+    if (this.currentActor(code) === seatId) {
+      const result = this.dispatch(code, seatId, "fold");
+      if ("steps" in result) fold = result;
```
ewanhardingham: **A rejected fold is swallowed entirely.** On rejection the seat is still freed and we're back to exactly bug #90 — hand waiting on a seat nobody owns.

It shouldn't be reachable: `currentActor` returns `toAct[0]`, and fold is unconditionally legal for the player to act. But there are two differences from the equivalent reasoning in `rescheduleActionClock`, which has a comment justifying the same assumption: here there's no comment, and there's no re-arm to recover if the assumption ever breaks.

Suggest either a short comment mirroring the `rescheduleActionClock` one, or routing the rejection through `logDispatch` so it shows up in the hand log rather than vanishing.

packages/server/src/rooms.ts
```diff
@@ -367,14 +372,22 @@ export class RoomStore {
...
    */
-  evictSeat(code: string, seatId: SeatId): void {
+  evictSeat(code: string, seatId: SeatId): EvictSeatResult {
     const room = this.#rooms.get(code);
     const seat = room?.seats[seatId];
-    if (!seat) return;
+    if (!seat) return {};
+
+    let fold: DispatchSuccess | undefined;
+    if (this.currentActor(code) === seatId) {
```
ewanhardingham: **Follow-up, out of #90's scope:** a *non-actor* mid-hand eviction still stalls the table for the full clock. I confirmed the behaviour — evicting a seat that's later in `toAct` returns no dispatch, leaves it in the hand, and it only resolves when the action clock fires. `DEFAULT_TIMEOUT_MS` is 90 s in production, so evicting two waiting seats costs three minutes of dead table.

#90 only specified the current actor, so shipping this as-is is defensible. But from the table device it reads as "evict → wait 90 seconds", which is the same family of bug. Worth filing rather than letting it disappear.

**Minor style, take or leave:** `EvictSeatResult` with one optional field plus `return fold === undefined ? {} : { dispatch: fold }` is a fair bit of `exactOptionalPropertyTypes` ceremony. Early-returning drops the `let`:

```ts
if (this.currentActor(code) !== seatId) {
  freeSeat(seat);
  return {};
}
const result = this.dispatch(code, seatId, "fold");
freeSeat(seat);
return "steps" in result ? { dispatch: result } : {};
```

The object return type itself is the right shape for growth.

packages/server/src/app.ts
```diff
@@ -417,6 +417,15 @@ export async function buildApp(
     });
   }
 
+  /** Publishes an accepted engine dispatch and moves the room's action clock. */
+  function publishDispatch(code: string, result: DispatchSuccess): void {
```
ewanhardingham, line 420: **`publishDispatch` silently drops `result.seatMoves`.** The WS caller applies them itself (`applySeatMoves`) just before calling this, and the helper ignores the field.

Harmless today — only `nextHand` produces moves and only the WS path dispatches it — but the name and the doc comment ("Publishes an accepted engine dispatch") promise more than the body delivers, and the next caller passing a move-bearing result loses them with no type error to catch it.

Either pull `applySeatMoves` inside the helper, or narrow the doc comment to state that seat moves stay the caller's responsibility.

packages/server/src/app.ts
```diff
@@ -417,6 +417,15 @@ export async function buildApp(
...
+    logDispatch(code, result);
+    for (const step of result.steps) {
+      fanOutHandUpdate(code, step);
+    }
+    rescheduleActionClock(code);
```
ewanhardingham, line 428: **Follow-up:** now that eviction folds go through `logDispatch`, the hand log records real actions and evict-folds but still not *clock-timeout* folds — `rescheduleActionClock` fans out its synthesized fold's steps without logging them.

Pre-existing, not caused by this PR, but this helper makes it a one-liner: in the timeout callback use `if ("steps" in result) publishDispatch(code, result);` and drop the trailing `rescheduleActionClock(code)` for that branch, since the helper re-arms. Worth doing here or filing.

packages/server/src/rooms.test.ts
```diff
@@ -635,6 +635,23 @@ describe("RoomStore", () => {
     });
 
     describe("ADR-0003: manual eviction", () => {
+      it("auto-folds the current actor before freeing the evicted seat", () => {
```
ewanhardingham, line 638: **Missing case: heads-up.** Both new tests use three seats. The 2-seat case takes a different engine path — the fold ends the hand instead of advancing the turn — and it's the one where `rescheduleActionClock` has to *clear* rather than re-arm.

I verified it works (`ActionTaken → HandFoldedOut → HandComplete`, `currentActor` undefined, clock cleared), so this is about pinning the behaviour, not fixing it. A near-copy of this test with 2 claimed seats asserting `store.currentActor(room.code)` is `undefined` and the hand status is `complete` would cover it.

packages/server/src/app.test.ts
```diff
@@ -1472,6 +1513,31 @@ describe("action clock", () => {
...
+    await opened(table.socket);
+    for (const seatId of [0, 1, 2]) rooms.claimSeat(room.code, seatId);
+    await settle();
+
+    table.socket.send(JSON.stringify({ type: "startHand" }));
+    await settle();
+    const actor = rooms.currentActor(room.code);
+    if (actor === undefined) throw new Error("expected a current actor");
+
+    await settle(120);
```
ewanhardingham: **Timing margins are ~50 ms on both sides.** `settle(120)` → evict → `settle(130)` against a 200 ms clock means the fold has to land inside a ~250 ms window, and the "wasn't reset" signal is that a reset clock would fire at ~325 ms > 250 ms.

It was stable across my runs, but under a loaded CI runner both edges are tight. `settle(150)` / `settle(120)` widens the pass side to ~70 ms without weakening the assertion (a reset clock would still fire at ~350 ms, well past the 270 ms check).

The test itself is a good guard — it's exactly what stops someone "simplifying" this into an unconditional reschedule on every eviction.

ewanhardingham (commented):
Correct, minimal, and well-targeted fix. Folding before `freeSeat` is the ordering that matters — `dispatch` builds the command from `seatId` and the engine has to still see a live actor — and only re-arming the clock when a fold actually happened is what makes the "different seat's eviction" regression test meaningful.

I fetched the branch and ran `packages/server/src/{app,rooms}.test.ts`: 110 passed, and four repeat runs of `app.test.ts` were stable. I also probed the heads-up case directly — evicting the actor in a 2-player hand emits `ActionTaken → HandFoldedOut → HandComplete`, `currentActor` goes `undefined`, and the clock clears. No crash, no stuck timer.

Inline comments below. The two I'd address before merge are the swallowed rejection and the `seatMoves` contract in `publishDispatch`; the rest are follow-ups.

ewanhardingham:
Addressed in commit 1ee9487: rejected eviction folds now leave the seat claimed for clock recovery; publishDispatch applies seat moves; timeout folds use the shared publisher and are logged; heads-up coverage and wider timing margins were added. The non-current-seat follow-up is tracked in issue #103.

PR #101, opened 2026-08-01
feat(table): show sitting-out seats as quiet pods (#92)

### LINKED ISSUE #92: Represent sat-out players visually at the table
## Problem

When a player sits out, the table interface currently has no visual treatment for their seat — an observer can't tell at a glance who is sitting out vs. actively in the hand.

## Ask

Design and implement a visual representation for sat-out seats at the table (e.g. dimmed/greyed seat, a "sitting out" badge, or similar), consistent with the tactile, non-basic UI direction of the frontend.

## Notes

- Should read clearly at a glance alongside other seat states (active, folded, empty, etc).
- Related recent work: start-time clock treatments for the hand picker (see commit 1a561ed on prototype/hand-picker-clock).

ewanhardingham:
## Summary

- promote Option A (Quiet pod) from the sat-out prototype into the production table UI
- keep the seat in its normal position, but mute it with a dashed pod/avatar, strike-through marker, and `Sitting out` / `Not in hand` copy
- add focused rendering coverage and shared theme tokens

## Design decision

Option A keeps the player easy to locate without making a non-participant compete with the active seats. The seat remains claimed and visually distinct from a folded or disconnected player.

Prototype source: `prototype/issue-92-sat-out` (`ccfa83d`)

Closes #92

## Validation

- `npm run build`
- `npm run lint`
- `npm run test --workspace @table-top-poker/table-client -- --run` — 57 tests passed
- `npm run build:app --workspace @table-top-poker/table-client`

### DIFF
--- a/packages/table-client/src/Seats.test.tsx
+++ b/packages/table-client/src/Seats.test.tsx
@@ -21,6 +21,15 @@ describe("Seats", () => {
     expect(html).toMatch(/data-testid="seat-pod-3"[^>]*data-status="open"/);
   });
 
+  it("gives a sitting-out seat a quiet local treatment and next-hand copy", () => {
+    const html = renderToStaticMarkup(<Seats seats={seats} view={null} />);
+
+    expect(html).toContain('data-testid="seat-pod-2-sitting-out"');
+    expect(html).toContain("Sitting out");
+    expect(html).toContain("Next hand");
+    expect(html).not.toContain('data-testid="seat-pod-0-sitting-out"');
+  });
+
   it("marks the button seat once a hand exists, even with no active betting", () => {
     const view: TableView = { phase: "no-hand", button: 1 };
     const html = renderToStaticMarkup(<Seats seats={seats} view={view} />);
--- a/packages/table-client/src/Seats.test.tsx
+++ b/packages/table-client/src/Seats.test.tsx
@@ -21,12 +21,40 @@ describe("Seats", () => {
     expect(html).toMatch(/data-testid="seat-pod-3"[^>]*data-status="open"/);
   });
 
-  it("gives a sitting-out seat a quiet local treatment and next-hand copy", () => {
+  it("gives a sitting-out seat a quiet local treatment and neutral copy", () => {
     const html = renderToStaticMarkup(<Seats seats={seats} view={null} />);
 
-    expect(html).toContain('data-testid="seat-pod-2-sitting-out"');
-    expect(html).toContain("Sitting out");
-    expect(html).toContain("Next hand");
+    expect(html).toMatch(
+      /data-testid="seat-pod-2-sitting-out"[^>]*>[\s\S]*Sitting out[\s\S]*Not in hand[\s\S]*<\/div>/,
+    );
+    expect(html).toContain('data-testid="seat-pod-2-sitting-out-marker"');
+    expect(html).toMatch(
+      /data-testid="seat-pod-2-avatar"[^>]*style="[^"]*border:1px dashed/,
+    );
+    expect(html).not.toContain('data-testid="seat-pod-0-sitting-out"');
+  });
+
+  it("shows a claimed seat absent from a live hand as sitting out", () => {
+    const view: TableView = {
+      phase: "betting",
+      button: 0,
+      street: "flop",
+      board: [],
+      toAct: [0],
+      seats: [
+        { seatId: 0, folded: false },
+        { seatId: 1, folded: false },
+      ],
+    };
+    const html = renderToStaticMarkup(<Seats seats={seats} view={view} />);
+
+    expect(html).toMatch(
+      /data-testid="seat-pod-2"[^>]*data-status="sitting-out"/,
+    );
+    expect(html).toMatch(
+      /data-testid="seat-pod-2-sitting-out"[^>]*>[\s\S]*Sitting out[\s\S]*Not in hand[\s\S]*<\/div>/,
+    );
+    expect(html).toContain('data-testid="seat-pod-2-sitting-out-marker"');
     expect(html).not.toContain('data-testid="seat-pod-0-sitting-out"');
   });
 
--- a/packages/table-client/src/Seats.test.tsx
+++ b/packages/table-client/src/Seats.test.tsx
@@ -167,11 +167,66 @@ describe("Seats", () => {
     expect(html).not.toContain("Winner");
   });
 
+  it("keeps a revealed player in-hand when they sit out for the next hand", () => {
+    const view: TableView = {
+      phase: "showdown",
+      button: 0,
+      board: [
+        { rank: "A", suit: "spades" },
+        { rank: "K", suit: "hearts" },
+        { rank: "2", suit: "clubs" },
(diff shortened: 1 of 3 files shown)


packages/table-client/src/Seats.tsx
```diff
@@ -190,15 +212,45 @@ export function Seats({ seats, view, onSeatClick }: SeatsProps) {
...
+                fontWeight: 700,
+                letterSpacing: "0.16em",
+                textTransform: "uppercase",
+                color: color.textMuted,
+              }}
+            >
+              Sitting out
+            </span>
+            <span style={{ fontSize: "0.6em", color: color.textFaint }}>
+              Next hand
```
ewanhardingham: **Blocking: `Next hand` is wrong for the voluntary sit-out case.**

`SeatView.sittingOut` is the server's *derived* flag (`isSittingOut`, `packages/server/src/rooms.ts:262`), which deliberately conflates two states:

1. a voluntary opt-out (the player pressed **Sit out**), and
2. a claim that arrived after the ring was fixed and hasn't been swept into a deal-in recompute yet.

Only (2) is dealt in next hand. For (1), `eligibleSeats` filters on `!seat.sittingOut` (`rooms.ts:232`), so the player stays out until they press **Sit in** — this pod promises them a return they won't get.

The two states aren't distinguishable client-side today, so I'd take the copy fix now (drop the second line, or use something state-neutral like `Not in hand`) and file the protocol split as a follow-up on #92.

packages/table-client/src/Seats.test.tsx
```diff
@@ -21,6 +21,15 @@ describe("Seats", () => {
     expect(html).toMatch(/data-testid="seat-pod-3"[^>]*data-status="open"/);
   });
 
+  it("gives a sitting-out seat a quiet local treatment and next-hand copy", () => {
+    const html = renderToStaticMarkup(<Seats seats={seats} view={null} />);
+
+    expect(html).toContain('data-testid="seat-pod-2-sitting-out"');
+    expect(html).toContain("Sitting out");
+    expect(html).toContain("Next hand");
```
ewanhardingham: **Blocking: the main production path is untested.**

This case runs with `view={null}`, which exercises the `else` branch of `deriveSeat` (`status = seat.sittingOut ? ... `, Seats.tsx:50). During a live hand the status comes from the *other* branch — `!handSeat` at Seats.tsx:45 — and that's where a sitting-out seat is actually seen most of the time. Please add a `phase: "betting"` case with a claimed seat absent from `view.seats`.

Two smaller things while you're here:
- `toContain("Sitting out")` / `toContain("Next hand")` aren't scoped to seat 2 — they'd pass if the copy leaked anywhere in the tree. The surrounding tests scope with `data-testid="seat-pod-N"[^>]*` regexes; worth matching that.
- Nothing asserts the quiet treatment itself (dashed ring, strike-through). For a pure-visual PR the strike-through span is cheap to pin down.

ewanhardingham (commented):
Two blocking items inline; the rest of my notes are in the summary review above.

packages/table-client/src/Seats.test.tsx
ewanhardingham: Addressed: added a betting-phase rendering case where claimed seat 2 is absent from `view.seats`; assertions scope the copy to that seat and pin both the dashed avatar ring and strike-through marker. The focused table suite now passes 57 tests.

packages/table-client/src/Seats.tsx
ewanhardingham: Addressed: changed `Next hand` to `Not in hand`. `SeatView.sittingOut` currently covers voluntary sit-out and the mid-hand-join case, so the copy is now neutral. The protocol distinction remains a follow-up to #92.

packages/table-client/src/Seats.tsx
```diff
@@ -251,15 +305,24 @@ export function Seats({ seats, view, onSeatClick }: SeatsProps) {
...
                 border: `1px solid ${
                   visual.isWinner
                     ? color.seatWinnerBorder
                     : visual.isActor
                       ? color.accent
-                      : "transparent"
+                      : visual.status === "sitting-out"
+                        ? color.seatSittingOutBorder
+                        : "transparent"
                 }`,
```
ewanhardingham, line 331: **A winning pod can be dimmed and struck through.** The `background` and `border` ternaries just above correctly let `isWinner` win, but this `opacity` chain checks `folded` then `sitting-out` with no winner branch — so a showdown winner who is flagged sitting-out renders at `0.82` with a faint, dashed, struck-through avatar next to their winning hole cards.

This is reachable through the normal UI, not a crafted-message edge case:

- `setSittingOut` (`packages/server/src/rooms.ts:439`) has no in-hand guard — it flips the flag immediately, whatever the engine is doing.
- The WS handler (`packages/server/src/app.ts:673-687`) routes `sitOut` straight through; ADR-0002 deliberately keeps it off the engine path.
- The player-side toggle is only disabled on disconnect (`StatusBar.tsx:57`), not mid-hand.

So: player presses **Sit out** during a hand → flag set, engine ring unchanged → they play to showdown and win. At `showdown` `deriveSeat` takes the else-branch (`Seats.tsx:50`) and reads the raw flag.

Note the `Next hand` copy is *correct* in this case — they really are out next hand. It's only the visual treatment that contradicts itself.

packages/table-client/src/Seats.tsx
```diff
@@ -190,15 +214,45 @@ export function Seats({ seats, view, onSeatClick }: SeatsProps) {
           </div>
         );
```
ewanhardingham, line 226: Related to my note on the `opacity` chain below, and the broader half of it: this isn't only a winner problem. *Any* seat that sat out mid-hand gets the full quiet treatment at showdown while its hole cards are face-up — dashed ring, strike-through, `Sitting out` — which reads as "not in this hand" about a player who visibly just played it.

Worth deciding explicitly rather than patching the winner case alone. The cleanest version is probably to gate the whole treatment on actually being out of the current hand — e.g. suppress it when `visual.holeCards` is non-null, or introduce a distinct status for "in this hand, out of the next one" so the pod can say something like `Out next hand` without borrowing the sat-out styling.

ewanhardingham (commented):
One more reachable issue, split across two comments — I confirmed the trigger path in the server rather than leaving it as a hypothetical.

packages/table-client/src/Seats.tsx
ewanhardingham: Addressed in 1a68fc0. The fix covers any revealed showdown participant, not only winners, by using `view.results` as the current-hand evidence; the folded-out winner is covered too. The protocol-level distinction for state-specific copy remains tracked in #104.

packages/table-client/src/Seats.tsx
ewanhardingham: Addressed in 1a68fc0. Completion-phase hand participation now takes priority over the room-level sitting-out flag: a showdown result or folded-out winner renders as in-hand, keeping opacity at 1 and removing the quiet copy, dashed ring, and strike-through marker. Added regression coverage for both paths.

Issue #88, opened 2026-07-31
Add player names to seats


ewanhardingham:
## Problem

Seats are currently only identified by seat number. Players have no way to attach a name to their seat, so both the table display and any dialog/log copy refer to "Seat 3" etc. instead of a person's name.

## Proposal

- When a player selects their seat (`packages/player-client/src/SeatPicker.tsx`), prompt for a display name as part of the seat-claim flow.
- Persist the name alongside the seat claim (engine `Seat`/`Room` state in `packages/engine/src/table.ts` / `room.ts`, and wherever seat tokens are stored — `packages/player-client/src/storage/seatToken.ts`).
- Render the name on the table view next to/under the seat (`packages/table-client`'s `Seats` component).
- Anywhere dialog or log copy currently references a seat by number, reference the player's name instead (falling back to "Seat N" if unnamed).

## Open questions

- Name validation/limits (length, empty, profanity) — TBD.
- What happens if two players pick the same name — TBD.
- Whether name is required to claim a seat or optional with a "Seat N" fallback — TBD.

## Scope

Touches `engine`, `player-client`, and `table-client`. Should be scoped/refined further (e.g. via wayfinder) before implementation if it turns out to span multiple decisions.

ewanhardingham:
Prototype captured on `prototype/issue-88-player-names` at [`776a370`](https://github.com/ewanhardingham/table-top-poker/commit/776a370).\n\nRun it from the prototype worktree with:\n`npm run dev --workspace @table-top-poker/player-client -- --host 127.0.0.1 --port 5199`\n\nOpen `http://127.0.0.1:5199/?prototype=issue-88&variant=A`; use the bottom switcher or `←` / `→` to compare:\n- A — inline name entry inside the selected seat\n- B — seat selection followed by a public-identity confirmation sheet\n- C — identity-first onboarding, then seat selection\n\nQuestion captured: where should identity enter the claim flow, and how should a name replace or fall back to the seat label in public copy? Initial read: A has the least interruption, B gives the clearest public-name confirmation, and C makes identity most central but adds the most setup friction. No winner has been promoted; the branch is the primary source for choosing one.

ewanhardingham:
[...]

PR #102, opened 2026-08-01
feat: add required player names to seats

ewanhardingham:
## Context

Issue #88 adds required player display names to seats while preserving the existing table layout. Names are trimmed, limited to 10 characters, and unique case-insensitively; profanity and public-name checks are intentionally out of scope.

The room seat count is mutable from 2 to 8. Before this change, an unclaimed player only received the initial room snapshot, so changing the table size left the seat picker stale and a race against a removed seat was reported as “someone just took that seat.”

## Implementation

- Persist display names through room views, reconnects, and positional seat repacks.
- Add larger, viewport-filling player seat cards with seat-count-aware rows.
- Add a room-scoped lobby WebSocket for unclaimed players; it receives room views but cannot issue table or player commands.
- Preserve truthful claim errors, including a specific message when a seat disappears during a claim.

## Validation

- `npm run lint`
- `npm run build`
- `npm test -- --reporter=dot --maxWorkers=1` — 67 files, 419 tests passing after merging main.
- Regression coverage verifies live seat-count growth and shrink updates (`4 → 6 → 2`) and rejects lobby commands.

### DIFF
--- a/packages/player-client/src/SeatPicker.tsx
+++ b/packages/player-client/src/SeatPicker.tsx
@@ -1,18 +1,45 @@
-import type { SeatView } from "@table-top-poker/protocol";
+import {
+  MAX_DISPLAY_NAME_LENGTH,
+  type SeatView,
+} from "@table-top-poker/protocol";
 import { color, font, fontSize, radius } from "@table-top-poker/ui-shared";
-import type { CSSProperties } from "react";
+import { useEffect, useState, type CSSProperties } from "react";
 import { InlineError } from "./InlineError.js";
 
 export interface SeatPickerProps {
   readonly seats: readonly SeatView[];
   readonly error: string | null;
   readonly evictionMessage?: string | null;
-  readonly onClaim: (seatId: number) => void;
+  readonly onClaim: (seatId: number, displayName: string) => void;
 }
 
 const seatErrorCopy: Record<string, string> = {
+  "room-not-found":
+    "That room is no longer available — return to the join screen.",
+  "seat-not-found": "That seat is no longer available — pick another.",
   "seat-already-claimed": "Someone just took that seat — pick another.",
+  "duplicate-display-name": "That name is already taken — choose another.",
+  "invalid-display-name": "Enter a name with 1–10 characters.",
+  "claim-failed": "Couldn't reach the table — try that seat again.",
 };
+
+/**
+ * Why a pending seat selection can no longer be claimed, or null while it
+ * still can. A live room view can take the seat or repack it out of the table
+ * between selecting it and claiming it.
+ */
+export function selectionLostMessage(
+  seatId: number,
+  seat: SeatView | undefined,
+): string | null {
+  if (seat === undefined) {
+    return `Seat ${String(seatId + 1)} is no longer at this table — pick another.`;
+  }
+  if (seat.claimed) {
+    return `Seat ${String(seatId + 1)} was just taken — pick another.`;
+  }
+  return null;
+}
 
 const descriptionStyle: CSSProperties = {
   fontSize: fontSize.md,
@@ -20,16 +47,26 @@ const descriptionStyle: CSSProperties = {
   color: color.textMuted,
 };
 
-const gridStyle: CSSProperties = {
-  display: "grid",
-  gridTemplateColumns: "1fr 1fr",
-  gap: 10,
-};
+function seatGridStyle(seatCount: number): CSSProperties {
+  const rowCount = Math.max(1, Math.ceil(seatCount / 2));
 
-function seatStyle(claimed: boolean): CSSProperties {
   return {
+    display: "grid",
+    gridTemplateColumns: seatCount <= 1 ? "1fr" : "repeat(2, minmax(0, 1fr))",
+    gridTemplateRows: `repeat(${String(rowCount)}, minmax(0, 1fr))`,
+    gap: 12,
+    flex: "1 1 0",
+    minHeight: 0,
+  };
+}
+
+function seatStyle(claimed: boolean, selected: boolean): CSSProperties {
+  return {
+    display: "flex",
+    minHeight: 108,
+    height: "100%",
     borderRadius: radius.control,
-    padding: "14px 13px",
+    padding: 16,
     ...(claimed
       ? {
           border: `1px solid ${color.border}`,
@@ -37,17 +74,27 @@ function seatStyle(claimed: boolean): CSSProperties {
           opacity: 0.5,
         }
       : {
-          border: `1px solid ${color.accentBorder}`,
+          border: `1px solid ${
+            selected ? color.accentBright : color.accentBorder
+          }`,
+          // Selection is carried by the brighter border and the ring, not by
+          // a fill: `lossBackground` is the error surface this same screen
+          // uses below, and a selected seat is not an error.
           background: color.accentWash,
+          boxShadow: selected ? `0 0 0 2px ${color.accentBorder}` : undefined,
         }),
   };
 }
 
 const rowStyle: CSSProperties = {
   display: "flex",
-  alignItems: "center",
-  gap: 11,
+  flexDirection: "column",
+  alignItems: "flex-start",
+  justifyContent: "space-between",
+  gap: 12,
   width: "100%",
+  height: "100%",
+  minHeight: 0,
   textAlign: "left",
   background: "none",
   border: 0,
@@ -59,8 +106,8 @@ const rowStyle: CSSProperties = {
 
 function avatarStyle(claimed: boolean): CSSProperties {
   return {
-    width: 34,
-    height: 34,
+    width: 44,
+    height: 44,
     borderRadius: radius.pill,
     flex: "none",
     display: "flex",
@@ -68,7 +115,7 @@ function avatarStyle(claimed: boolean): CSSProperties {
     justifyContent: "center",
     fontFamily: font.mono,
     fontWeight: 700,
-    fontSize: fontSize.xs,
+    fontSize: fontSize.sm,
     background: color.avatarGradient,
     color: color.pillInk,
     filter: claimed ? "saturate(.2) brightness(.7)" : undefined,
@@ -79,6 +126,7 @@ const textColStyle: CSSProperties = {
   display: "flex",
   flexDirection: "column",
   gap: 2,
+  width: "100%",
   minWidth: 0,
 };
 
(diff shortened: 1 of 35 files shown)


packages/player-client/src/ws/useLobbyWebSocket.ts
```diff
@@ -0,0 +1,73 @@
...
+        getWebSocketUrl(window.location, { room: code, role: "lobby" }),
+      );
+      socketRef = socket;
+
+      socket.addEventListener("open", () => {
+        openedOnce = true;
+      });
+      socket.addEventListener("close", () => {
+        if (!active) return;
+        if (!openedOnce) {
```
ewanhardingham, line 61: Closed-before-open is treated as "the room ended", which in `App` clears the room, clears the stored seat token, and drops the player back to the join screen. But the same condition fires for a transient failure — server restart, dev proxy hiccup, wifi drop at exactly the wrong moment — where nothing about the room has actually changed.

`useWebSocket` makes this distinction explicit: it maps closed-before-open to `onRejected` ("an upgrade-time rejection, not a network blip") because a stale seat token would otherwise spin forever. The lobby socket has no such terminal case — the only upgrade rejections are 400 `room-required` and 404 `room-not-found`, and a genuinely-ended room also sends `room-ended` on the wire before closing.

Suggest retrying on close regardless, and letting the explicit `room-ended` message (already handled below) be the only path to `onRoomEnded`. If you want to keep the terminal case, re-checking `GET /rooms/:code` before giving up would distinguish the two.

packages/player-client/src/claimError.ts
```diff
@@ -0,0 +1,21 @@
...
+  if (
+    code === "room-not-found" ||
+    code === "seat-not-found" ||
+    code === "seat-already-claimed" ||
+    code === "invalid-display-name" ||
+    code === "duplicate-display-name"
+  ) {
+    return code;
+  }
+  return "seat-already-claimed";
```
ewanhardingham: The PR description says claim errors are now truthful, but this fallback still makes every non-enumerated failure claim someone took the seat. `claimSeat` throws a bare `Error` for a fetch-level network failure and for `failed to claim seat: 500`; both surface as "Someone just took that seat — pick another", which sends the player hunting for a different seat when the seat was fine and the server wasn't.

Consider returning a distinct `"claim-failed"` code for the fallback with copy like "Couldn't reach the table — try again", keeping `seat-already-claimed` for the actual 409.

packages/server/src/rooms.ts
```diff
@@ -354,15 +365,36 @@ export class RoomStore {
     this.#rooms.delete(code);
   }
 
-  claimSeat(code: string, seatId: SeatId): ClaimSeatResult {
+  claimSeat(
+    code: string,
+    seatId: SeatId,
+    displayName?: string,
```
ewanhardingham: `displayName` is optional here, so `RoomStore` still permits a nameless claim, and every validation rule (non-empty, length, uniqueness) is skipped entirely when it's `undefined`. The "names are required" invariant therefore lives only in the HTTP route's `ClaimSeatRequestSchema`, and the store — the thing that owns the seat — can't enforce it.

Since the only caller now always passes a name, making the parameter required (`displayName: string`) costs nothing, deletes the `trimmedName !== undefined` branch, and makes the type say what issue #88 asks for. The optional field on `Seat` should stay for pre-name rooms; that's a different concern from what a *new* claim is allowed to do.

packages/server/src/app.ts
```diff
@@ -378,7 +388,7 @@ export async function buildApp(
           event: redactEventFor(step.event, "table"),
           view: view(step.state, "table"),
         });
-      } else if (step.state.seats.includes(identity)) {
+      } else if (identity !== "lobby" && step.state.seats.includes(identity)) {
```
ewanhardingham: This guard and its twin in `broadcastDisplayedHand` are what stop an unauthenticated lobby socket from receiving `view(state, seatId)` — the hole-card path. They're correct, but they're the security boundary of this PR and neither is covered by a test. The new lobby test only asserts seat-count pushes and `not-permitted` on `startHand`.

Worth adding: start a hand with two claimed seats, keep a `role=lobby` socket open, and assert it received only `room-view` messages and never a `hand-update` or `view-snapshot`.

Separately, `identity !== "lobby"` now appears ~10 times across this file, each one an independent chance to forget. A single `function isSeat(identity: SocketIdentity | undefined): identity is SeatId` would narrow correctly at every call site and make a missed check a type error rather than a silent leak.

packages/server/src/app.ts
```diff
@@ -587,7 +606,7 @@ export async function buildApp(
             await reply.code(404).send({ error: "room-not-found" });
             return;
           }
-          if (role === "table") return;
+          if (role === "table" || role === "lobby") return;
```
ewanhardingham: `role` is an unvalidated free-form string, so this reads as "anything that isn't `table` or `lobby` must authenticate" — fine today, but a typo'd role (`role=lobbby`) silently falls through to seat auth and 403s rather than failing loudly. Validating `role` with `z.enum(["table", "lobby"]).optional()` in the query schema would keep the set of unauthenticated roles explicit and reviewable in one place.

packages/table-client/src/App.tsx
```diff
@@ -33,8 +34,23 @@ export function App() {
...
   const clearRoom = useTableStore((state) => state.clearRoom);
 
+  useEffect(() => {
+    if (typeof window === "undefined") return;
+    const code = new URLSearchParams(window.location.search)
+      .get("room")
+      ?.trim()
+      .toUpperCase();
+    if (!code) return;
+    joinRoom(code)
```
ewanhardingham: This is unannounced scope — nothing in the PR description or issue #88 mentions a table-client deep link, and it isn't covered by a test.

More importantly, `setRoomView` sets `roomCode`, which is what `useWebSocket` keys on, so this URL opens a `role=table` socket. `role=table` bypasses seat-token auth, so anyone who can guess or read a 4-character room code now gets a full table device by URL — including the evict, end-session, and seat-count controls, not just a read-only view. That exposure technically predates this change, but until now it required hand-crafting a WebSocket; this makes it a link.

Also: `console.error` is the only failure handling, so a bad or expired code leaves the table on a blank screen with no explanation, and the code isn't validated before it's sent.

I'd split this into its own issue and land it with an auth story (or at minimum a read-only role) rather than riding along with player names.

packages/player-client/src/SeatPicker.tsx
```diff
@@ -102,30 +170,50 @@ export function SeatPicker({
   evictionMessage,
   onClaim,
 }: SeatPickerProps) {
+  const [selectedSeatId, setSelectedSeatId] = useState<number | null>(null);
+  const [displayName, setDisplayName] = useState("");
+  const selectedSeat =
+    selectedSeatId === null
+      ? undefined
+      : seats.find((seat) => seat.id === selectedSeatId);
+  const canClaim = selectedSeat !== undefined && displayName.trim() !== "";
```
ewanhardingham: `selectedSeatId` isn't reconciled against incoming room views — which now arrive live over the lobby socket, so this happens more often than it used to. If another player claims the selected seat, `selectedSeat.claimed` flips true, the name-entry panel unmounts, and whatever the player had typed disappears with no message; the card just quietly reads "Taken". Same silent drop when the table shrinks the seat count out from under the selection.

Clearing `selectedSeatId` in an effect when the seat becomes claimed or vanishes, and setting an inline notice ("Seat 3 was just taken — pick another"), would keep the picker honest under exactly the race this PR set out to fix.

packages/player-client/src/SeatPicker.tsx
```diff
@@ -20,34 +28,51 @@ const descriptionStyle: CSSProperties = {
...
           background: color.mutedSurface,
           opacity: 0.5,
         }
       : {
-          border: `1px solid ${color.accentBorder}`,
-          background: color.accentWash,
+          border: `1px solid ${
+            selected ? color.accentBright : color.accentBorder
+          }`,
+          background: selected ? color.lossBackground : color.accentWash,
```
ewanhardingham: `color.lossBackground` for the *selected* seat is a semantic mismatch — the same token is used at line 303 for the error box, so a selected seat and a failed claim end up wearing the same colour. If there's no dedicated selection token in `ui-shared`, `color.accentWash` with the existing `boxShadow` ring probably carries the state on its own; otherwise adding a `selectedBackground` token would be worth it.

packages/player-client/src/SeatPicker.tsx
```diff
@@ -102,30 +170,50 @@ export function SeatPicker({
...
       }}
     >
       <div style={descriptionStyle}>
         Pick where you're sitting. Seat order sets the button and blinds.
       </div>
-      <div style={gridStyle}>
+      <div
+        data-testid="seat-grid"
+        data-seat-count={seats.length}
+        data-seat-columns={seats.length <= 1 ? 1 : 2}
```
ewanhardingham: `data-seat-columns` and `data-seat-rows` duplicate values already expressed in `seatGridStyle`, and exist only so the test can assert them — the test then also asserts the rendered `grid-template-rows`, so they aren't even load-bearing. Dropping both keeps the derivation in one place; `data-seat-count` is arguably worth keeping as a stable hook, the other two are shipping test scaffolding to production DOM.

packages/table-client/src/HouseRulesSheet.tsx
```diff
@@ -262,7 +265,7 @@ export function HouseRulesSheet({
                   {moves
                     .map(
                       (move) =>
-                        `${seatLabel(move.from)}→${String(move.to + 1)}`,
+                        `${seatLabel(move.from, seats)}→${String(move.to + 1)}`,
```
ewanhardingham: The two halves of the arrow no longer describe the same kind of thing — `Blair→2` reads as a name mapping to a number. Suggest `${seatLabel(move.from, seats)} → Seat ${move.to + 1}` so the destination stays labelled, which also keeps the string legible when a name happens to be numeric ("7→2").

ewanhardingham (commented):
Review of the display-name work. The server-side story is solid — `SeatView.displayName` is optional so old rooms still render, `freeSeat` clears the name on eviction, and `repackSeats` carries it through a shrink. The lobby socket is correctly excluded from every per-seat view path, so no hole cards leak to an unauthenticated connection.

Main things I'd want addressed before merge:

1. **The lobby socket's card-redaction guards are untested.** `app.ts` now has ~10 separate `identity !== "lobby"` checks; the only lobby tests cover seat-count pushes and command rejection. The one that actually matters — that a lobby socket receives no `hand-update` / `view-snapshot` while a hand is live — has no test. Adding one, plus collapsing the checks behind a single `isSeatIdentity(identity)` predicate, would make the invariant hard to break later.
2. **`?room=` auto-join on the table client is out of the stated scope** and turns any browser with a room code into a full table device (evict, end session, seat count) with no auth and no error surfaced to the user.
3. **A network blip on the lobby socket's first connect is reported as "the room ended"**, discarding the room and the stored seat token.

Rest is inline. Nothing here is blocking correctness of the naming feature itself.

Issue #143, opened 2026-08-01
feat(player-client): cancellation, resets and inert-by-default server views


ewanhardingham:
## Parent

Part of #138.

## What to build

An interrupted gesture always lands somewhere stable, and **nothing the table does yanks the cards out of the Player's hands**.

- **Cancellation** — pointer cancel, lost capture, hidden page — returns the pair to its last stable state and clears the recognizer to `Idle`: `Peeking → FaceDown`; below-threshold `FoldDragging → FaceDown` or `Revealed`, whichever it started from (§10).
- **`Turning` is a point of no return** for pointer lift and pointer cancellation — the flip completes. Cancellation must restore *stable* presentation, and mid-turn is not stable, so completing the turn **is** the restoration (§10).
- **Resets to face-down**, cancelling any active gesture: a new Hand, reload or reconnect, and the app leaving the foreground (`visibilitychange → hidden`, subscribed by the hook as a document event, not a prop change) (§9).
- A hand-boundary reset **snaps** face-down without animating, including out of `Turning` — no face-up frame of the previous Hand may survive into the next (§10).
- **Explicitly not resets**: ordinary turns, street changes, other Players' actions, and **the Player's own Actions**. Pressing Call, Raise or Check leaves the cards exactly as they were (§9, story 45).
- **Server views are inert by default** (§6). This ticket is where that is proven: everything except the three named disturbances is presentation-invisible and must not touch the recognizer.

## Acceptance criteria

- [ ] `CANCELLED` returns to the last stable state from `Peeking`, from below-threshold `FoldDragging` starting face-down, and from below-threshold `FoldDragging` starting revealed
- [ ] `CANCELLED` during `Turning` does **not** interrupt; the flip completes
- [ ] `RESET` during `Turning` snaps to `FaceDown` without animating
- [ ] `RESET` cancels an active gesture and lands on `FaceDown` from **every** presentation state
- [ ] `visibilitychange → hidden` produces `RESET`; cards are face-down on return from the background, and after a reload or reconnect
- [ ] `viewEvents.test.ts` asserts the **absence** of events: a new street, a new board card, another Player's bet, a changed `toAct`, and an unchanged-cards re-render each produce `[]`
- [ ] A peek in progress survives an incoming server update
- [ ] Pressing Call, Raise or Check leaves presentation untouched

## Conflict notes

[...]

Issue #141, opened 2026-08-01
feat(player-client): bend the corner to peek at your Hole cards


ewanhardingham:
## Parent

Part of #138.

## What to build

The Player drags the bend affordance on the bottom-right corner of the pair and both cards **peel open together**, tracking the finger. Lifting the finger closes them immediately. A glance costs nothing and leaves nothing exposed.

This is the chokepoint ticket: it brings in the recognizer that every remaining gesture builds on.

- The Recognizer machine — `Idle`, `Pressing`, `Bending`, `FoldDragging` (with its `armed` flag), `Ignored`, `Committed` — reduced by the **same single pure `reduce`** as Presentation, so a coupled event applies atomically (§3).
- `classify()` as a separate pure function, run **exactly once** on the first move past 9px of slop and never revisited (§4). Stickiness is a property of the reducer: the classification event is only accepted from `Pressing`.
- `Peeking(p)` presentation, `p ∈ [0,1)`, closing on release.
- Bend progress is `max(0, -dx) + max(0, -dy)` — leftward and upward at equal weight — so a pure leftward drag drives the peel at full rate and cannot be stolen by the fold recognizer (§15).
- Motion stays off React's render path: continuous values are `MotionValue`s in refs, written by pointer handlers and read by `BendableCard`. The reducer receives threshold **crossings** as events, never per-frame progress (§13).
- The two in-gesture bend prompts (§11 copy table): leftward-dominant, and upward-dominant with its "drag left to keep your finger clear" second line. These are **permanent** for every Player and never retire.
- Platform hardening (§16): `user-select: none`, `-webkit-touch-callout: none`, `touch-action` set so the browser does not claim the vertical drag, and no double-tap zoom.

**Multi-touch: first pointer wins.** While a gesture is active, any additional `pointerdown` is ignored until the first pointer releases or cancels.

## Acceptance criteria

- [ ] `classify.test.ts` covers the §4 table: bend zone wins while face-down; bend zone does **not** classify as `Bending` when already revealed; upward-dominant only past the 1.05 ratio; `foldLegal: false` falls through to `Ignored`; sideways and downward both `Ignored`
- [ ] A second classification event is a no-op — accepted only from `Pressing`
- [ ] `Ignored` is terminal until release and does not tap on release
- [ ] A drag under 9px of slop is not classified at all, so a tap with a wobble still reads as a tap
- [ ] A second `pointerdown` mid-gesture is ignored
[...]

Issue #138, opened 2026-08-01
spec(player-client): tactile Hole-card interactions


ewanhardingham:
Part of #107

Build-ready Phase 3 specification for tactile Hole-card interactions on the Player's phone. It synthesises the five closed map tickets — the grammar (#108), the accessibility constraints (#109), the lifecycle contract (#110), the module seam (#111) and the coaching copy (#135) — into one implementable whole, and settles the two questions that fell between them.

## Problem Statement

A Player's Hole cards are the only genuinely private thing in a room where everything else is shared on the table device. Today the player client just prints them face-up and leaves them there: they arrive at deal time, they render as two static face-up cards, and they stay face-up for the whole Hand.

That is wrong in three ways that matter at a real table.

**It is not private.** Face-up cards for the entire Hand means a phone lying on a table, held at a bad angle, or glanced at by the person beside you leaks the one secret the game is built on. At a physical table nobody leaves their cards face-up for ten minutes; they squeeze a corner, look, and put them back down. The client offers no equivalent, so the Player has to manage privacy by hiding the whole phone.

**It does not feel like poker.** Handling cards is most of the physical pleasure of the game — the squeeze, the peel, the flick into the muck. The player client is a screen with two pictures on it and four buttons underneath. Every Action is a form submission. Nothing about it rewards the hand that is already holding the device.

**Fold is the same weight as everything else.** Fold is the one irreversible, money-losing Action, and it is a button in a grid of four identical buttons, one thumb-slip away from Check. There is no physical commitment to making it and no moment where the Player can feel themselves about to.

## Solution

The Hole cards become an object the Player handles, not a picture they read.

A new Hand deals them **face-down**, overlapped, with a visible bend affordance on the bottom-right corner of both. Nothing is exposed until the Player asks for it.

**Bending that corner peeks.** Drag the corner and the pair peels open together — a temporary Hole-card peek that closes the moment you let go, exactly like lifting the edge of two physical cards. Drag past 90% of the peel and you commit: both cards turn face-up and **stay** face-up. That is a Hole-card reveal — persistent, local, and entirely private. Neither peek nor reveal touches the server, changes poker visibility, or affects Showdown.

**A tap conceals.** One tap on a revealed pair puts both cards face-down immediately. No animation, no delay — concealment is a privacy act and privacy acts should not have a wind-up.

[...]

Issue #145, opened 2026-08-01
feat(player-client): swipe the Hole cards up to Fold


ewanhardingham:
## Parent

Part of #138.

## What to build

The Player swipes the pair **upward, away from them**. The cards track the finger. Past a threshold the interaction **arms and says so**; release there and the cards fly away into a temporary muck and the Fold is sent. Release short of it and they drop safely back.

**The commitment is on release, never on crossing the line** — the Player can always change their mind by putting the cards down (§10).

- `FoldDragging` carrying an `armed` flag, classified per §4: `foldLegal && dy < 0 && |dy| > |dx| * 1.05`.
- Fold threshold `max(148px, 0.18 × viewport height)` (§15).
- **The arming signal is card motion plus permanent in-gesture text** (§11): "Keep dragging up" below the threshold, "Release to Fold" once armed. There is **no rendered fold-threshold marker**, and haptics are Blink-only — so on iPhone/Safari the text *is* the arming signal. It never retires, for any Player, ever.
- An optional best-effort haptic pulse at the threshold where the browser supports it. **Never semantic feedback**; the iPhone/Safari path must work with none (§16).
- On commit: presentation → `Leaving`, Fold sent through `useActionIntent`. The flight is **fire-and-forget** on its own ~280ms schedule, not gated on the round trip (§7).
- **Acknowledged → `Absent`** (visually a no-op if the flight has finished). **Rejected → `FaceDown`**, using the existing rejection message, and **the return interrupts the outbound flight immediately**, reversing from wherever the cards are rather than completing the departure first. Never back to `Revealed`.
- **The pair is inert while pending** — no bend, no tap, no drag. There is no undo once the server accepts.
- **The cards leave with whatever face they had**: a revealed pair flies away face-up.
- The Fold button in `ActionBar` is unchanged and remains available — the non-dragging path for a shaky hand, an awkward device, or assistive technology (§12).

**Accepted costs, intentional:** a fold cannot be started from the bend zone while face-down (§4), and a revealed pair flies to the muck face-up (§7).

## Acceptance criteria

- [ ] Release above the threshold commits Fold; release below returns to `FaceDown` or `Revealed`, whichever it started from
- [ ] Crossing the threshold alone commits nothing — only the completing pointer release does
- [ ] A drag that starts sideways or downward stays `Ignored` and cannot become a fold by curving upward
- [ ] `FaceDown | Revealed → Leaving` on commit; `Leaving → Absent` on acknowledgement
[...]

PR #155, opened 2026-08-02
feat(player-client): double-tap the Hole cards to Check

### LINKED ISSUE #144: feat(player-client): double-tap the Hole cards to Check
## Parent

Part of #138.

## What to build

Two taps on the pair sends **Check, and only ever Check** — never Call, never anything off-turn, never when Check is illegal. The commonest free Action becomes a reflex rather than a target.

- **No double-tap arbitration timer** (§5). Check fires on the second tap's *arrival*, not on a timer expiring, so both actions get faster.
- **Face-down:** first tap is a no-op; a second tap inside the 280ms window sends Check.
- **Revealed:** first tap conceals **immediately**; a second tap inside the window sends Check on top of the concealment. This reads as "I'm done looking, I check", and the concealment lands visibly before the Check confirmation.
- A visible confirmation when a gesture Check lands, so the Player knows the Action went without watching the buttons (story 31).
- The gesture enters `actions.check`, which **is** `intent.check` — `canAct` remains the single legality gate (§2). `checkLegal` is for arming and rendering only; a stale prop can at worst arm a gesture `canAct` then refuses.
- The Check button in `ActionBar` is unchanged and remains available.

**Accepted cost, stated in the spec so it is not rediscovered as a bug:** Check cannot be sent while keeping the cards face-up. One tap gets them back.

## Acceptance criteria

- [ ] A double-tap sends `check` and never `call`
- [ ] A double-tap off-turn is a no-op
- [ ] A double-tap when Check is illegal is a no-op
- [ ] From `Revealed`, the first tap conceals and the second sends Check
- [ ] From `FaceDown`, the first tap is a no-op and the second sends Check
- [ ] Taps outside the 280ms window do not compose into a Check
- [ ] Gesture and button reach `intent.check` by the identical route — no parallel path
- [ ] A visible confirmation renders when a gesture Check lands
- [ ] `ActionBar` and `useActionIntent` are unmodified

## Conflict notes

Extends the tap path landed by #142 in `useHoleCards.ts` and adds arms to `cardState.ts` — shared with the cancellation and Fold tickets.

The Check confirmation surface and #145's muck-flight styling both land presentational chrome in `HoleCardPair.tsx`; neither is prescribed by the spec, so align both with the app's current theme tokens.

## Blocked by

- #141 — bend to peek.

ewanhardingham:
Closes #144.

## What this does

A second tap inside 280ms sends Check the moment it lands — no timer arbitrating whether a single tap was still coming. A tap on a revealed pair conceals it instantly, so the conceal is visible before the Check it may turn out to be half of. The accepted trade-off from the issue holds: a Check cannot be sent while keeping the cards face-up, and one tap gets them back.

`actions.check` **is** `intent.check`, so the gesture and the ActionBar button reach the Action by the identical route and `canAct` — on the latest view, inside the intent — stays the single legality gate. An off-turn or illegal double-tap is inert without a guard of its own. `checkLegal`/`pending` are read only to decide whether to *claim* the Action landed, exactly as `ports.ts` promises.

`ActionBar` and `useActionIntent` are untouched; every file in the diff is inside `holecards/`.

## Shape of the change

| File | Why |
| --- | --- |
| `taps.ts` (new) | The double-tap window as a pure value, plus the rendering-only "should this confirm?" predicate |
| `cardState.ts` | The `TAPPED` and `DOUBLE_TAPPED` arms the module had declared and left unanswered |
| `coaching.ts` | The Check confirmation, as the one hint that is news rather than advice |
| `useHoleCards.ts` | Binds the tap to the Action and holds the transient confirmation |
| `constants.ts` | `CHECK_CONFIRM_MS`, carried forward from the prototype |

## One deliberate deviation, flagged for review

#138 §12 guarantees "Enter, Space **or a click** toggles reveal and conceal". To make #144's *"face-down state: first tap is a no-op"* true, a completed pointer gesture now disowns its synthetic click whether or not it classified — otherwise the click answers the tap a second time, re-revealing the pair the tap just concealed.

The consequence: **a desktop mouse click on a face-down pair no longer reveals it.** Enter and Space still do (they raise a click with no pointer sequence in front of it), a mouse drag on the corner still bends, and a mouse double-click now sends Check. The two spec statements can't both hold; I took #144's explicit acceptance criterion as the more recent word. Say if you'd rather keep the click and weaken the criterion instead.

## Testing

Pure-module contract tests only — `taps`, the two new reducer arms, and the confirmation's place in the hint ranking. The repo has no DOM test environment (`environment: "node"`, no jsdom), so the hook wiring itself is verified by construction, as it was for #141 and #143. This is the first slice where an Action leaves the module on a gesture, so it may be worth deciding separately whether the component-level layer #138's testing strategy names should exist.

Full suite: 581 passed. `tsc -b`, `npm run lint`, `npm run build` all clean.

https://claude.ai/code/session_01P2ongHEHMVVcoRkYJLHGqm

### DIFF
--- a/packages/player-client/src/holecards/cardState.ts
+++ b/packages/player-client/src/holecards/cardState.ts
@@ -14,8 +14,8 @@
  *
  * Answered so far: the deal-in, the emptying and the keyboard reveal/conceal
  * toggle (#139), the showdown reveal and lock (#140), the press, the
- * classification and the release that make up a bend (#141), and cancellation
- * and resets (#143).
+ * classification and the release that make up a bend (#141), cancellation
+ * and resets (#143), and the tap and double-tap that conceal and Check (#144).
  */
 
 import type { Classification } from "./classify.js";
@@ -278,13 +278,34 @@ export function reduce(state: CardState, event: CardEvent): CardState {
       if (state.presentation === "Absent") return state;
       return idle("FaceDown");
 
+    case "TAPPED":
+      // One tap hides the hand the moment someone leans over (story 5), and
+      // does nothing at all to a pair that is already face-down — which is
+      // what makes the first tap of a double-tap Check free (§5). A revealing
+      // tap would put a face-up frame between the two taps and charge the
+      // commonest free Action a reveal.
+      //
+      // Mid-turn it is dropped rather than queued, for the same reason
+      // `ACTIVATED` is: the flip is a point of no return.
+      return state.presentation === "Revealed"
+        ? { ...state, presentation: "FaceDown" }
+        : state;
+
+    case "DOUBLE_TAPPED":
+      // Presentation-wise, nothing: the first tap already concealed, and the
+      // second one buys a Check rather than a card movement. The Action itself
+      // is an **effect**, which this function cannot have — the hook sends it
+      // through `actions.check`, so a gesture and the button reach
+      // `intent.check` by the identical route and `canAct` stays the single
+      // legality gate (§2). Reducing it here anyway is what makes the event
+      // inert against a decided hand for free, through `survivesLock`.
+      return state;
+
     // Declared in the union so the shape is settled up front, and answered by
-    // the slices that own them: tap-conceal (#142), the double-tap Check
-    // (#144), the fold drag (#145) and fold disarming (#146).
+    // the slices that own them: the fold drag (#145) and fold disarming
+    // (#146).
     case "FOLD_ARMED":
     case "FOLD_DISARMED":
-    case "TAPPED":
-    case "DOUBLE_TAPPED":
     case "PENDING_RESOLVED":
       return state;
 
--- a/packages/player-client/src/holecards/coaching.ts
+++ b/packages/player-client/src/holecards/coaching.ts
@@ -18,6 +18,13 @@ export interface Hint {
   readonly id: string;
   readonly line1: string;
   readonly line2: string | null;
+  /**
+   * Whether the hint is news rather than advice, and should therefore reach a
+   * screen reader when it appears. Only the Check confirmation is: every
+   * teaching hint and in-gesture prompt describes what the player *could* do,
+   * and announcing those over a live gesture would be noise.
+   */
+  readonly announce: boolean;
 }
 
 export interface HintContext {
@@ -27,6 +34,8 @@ export interface HintContext {
   readonly locked: boolean;
   /** ~2s since the last contact with the pair. */
   readonly quiet: boolean;
+  /** A gesture Check landed moments ago and is still being confirmed (§5). */
+  readonly checkConfirmed: boolean;
 }
 
 /**
@@ -55,6 +64,12 @@ export function selectHint(
   _discovered: ReadonlySet<TeachableGesture>,
   ctx: HintContext,
 ): Hint | null {
+  // Outranks every gate below, including the pending one it necessarily trips:
+  // this is not advice about a gesture, it is the answer to one the player just
+  // made (story 31). Without it the only sign a double-tap landed is the
+  // ActionBar the gesture exists to stop them watching.
+  if (ctx.checkConfirmed) return checkedHint;
+
   // A hint is only ever advice about cards the player is actually holding, in
   // a moment where those cards will listen.
   if (ctx.pending || ctx.locked) return null;
@@ -70,10 +85,24 @@ export function selectHint(
   return null;
 }
 
+/**
+ * The visible confirmation a gesture Check lands with (story 31). It says the
+ * Action, not the gesture, because that is what the player needs to trust: the
+ * cards are already back face-down by the time it appears, so nothing else on
+ * this surface has changed to say the double-tap was heard.
+ */
+const checkedHint: Hint = {
+  id: "checked",
+  line1: "Checked",
+  line2: "your turn passed on",
+  announce: true,
+};
+
 const bendLeftHint: Hint = {
   id: "bending-left",
   line1: "Release to peek",
   line2: "keep bending to reveal both",
+  announce: false,
 };
 
 /**
@@ -85,4 +114,5 @@ const bendUpHint: Hint = {
   id: "bending-up",
   line1: "Release to peek",
   line2: "drag left to keep your finger clear",
+  announce: false,
 };
--- a/packages/player-client/src/holecards/taps.test.ts
+++ b/packages/player-client/src/holecards/taps.test.ts
@@ -1,0 +1,71 @@
+import { describe, expect, it } from "vitest";
+import { DOUBLE_TAP_MS } from "./constants.js";
+import { confirmsCheck, tapLanded } from "./taps.js";
+
+describe("tapLanded", () => {
+  it("opens the window on the first tap", () => {
+    expect(tapLanded(null, 1000)).toEqual({
+      window: 1000,
+      event: { type: "TAPPED" },
+    });
+  });
+
+  it("composes a Check from a second tap inside the window", () => {
+    const first = tapLanded(null, 1000);
+
+    expect(tapLanded(first.window, 1000 + DOUBLE_TAP_MS - 1)).toEqual({
+      window: null,
+      event: { type: "DOUBLE_TAPPED" },
+    });
+  });
+
+  it("counts a tap landing exactly on the window's edge", () => {
+    const first = tapLanded(null, 1000);
+
+    expect(tapLanded(first.window, 1000 + DOUBLE_TAP_MS).event).toEqual({
+      type: "DOUBLE_TAPPED",
+    });
+  });
+
+  it("does not compose a Check from taps outside the window", () => {
+    const first = tapLanded(null, 1000);
+
+    expect(tapLanded(first.window, 1000 + DOUBLE_TAP_MS + 1)).toEqual({
+      window: 1000 + DOUBLE_TAP_MS + 1,
+      event: { type: "TAPPED" },
+    });
+  });
+
+  it("closes the window on the Check, so a third tap starts a fresh pair", () => {
+    const first = tapLanded(null, 1000);
+    const second = tapLanded(first.window, 1100);
+    const third = tapLanded(second.window, 1200);
+
+    expect(second.event).toEqual({ type: "DOUBLE_TAPPED" });
+    // Without this, a triple-tap would send two Checks 100ms apart.
+    expect(third).toEqual({ window: 1200, event: { type: "TAPPED" } });
+  });
+
+  it("keeps a slow stream of taps composing nothing", () => {
+    let window = tapLanded(null, 0).window;
+    for (const now of [400, 800, 1200]) {
+      const step = tapLanded(window, now);
+      expect(step.event).toEqual({ type: "TAPPED" });
+      window = step.window;
+    }
+  });
+});
+
+describe("confirmsCheck", () => {
+  it("confirms a Check the player could actually make", () => {
+    expect(confirmsCheck({ checkLegal: true, pending: false })).toBe(true);
+  });
+
+  it("says nothing when Check is not on offer — off-turn, or facing a bet", () => {
+    expect(confirmsCheck({ checkLegal: false, pending: false })).toBe(false);
+  });
+
+  it("says nothing while another Action is already in flight", () => {
+    expect(confirmsCheck({ checkLegal: true, pending: true })).toBe(false);
+  });
+});
--- a/packages/player-client/src/holecards/taps.ts
+++ b/packages/player-client/src/holecards/taps.ts
@@ -1,0 +1,62 @@
+import type { CardEvent } from "./cardState.js";
+import { DOUBLE_TAP_MS } from "./constants.js";
+import type { CardActions } from "./ports.js";
+
+/**
+ * The double-tap window as a value (Phase 3 spec #138 §5, story 27).
+ *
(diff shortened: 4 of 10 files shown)


packages/player-client/src/holecards/useHoleCards.ts
```diff
@@ -156,28 +176,76 @@ export function useHoleCards(props: HoleCardPairProps): HoleCards {
...
-    if (active.classification !== null) {
-      disownClicksUntil.current = Date.now() + CLICK_DISOWN_MS;
-    }
+    // Every completed gesture disowns its click, a tap included: the tap is
+    // answered here, by the recognizer, and the click the browser synthesises
+    // afterwards would answer it a second time — re-revealing the pair the tap
+    // just concealed, or revealing on the first tap of a Check that §5 requires
+    // to cost nothing. §12's non-gesture path is unaffected: Enter and Space
+    // raise a click with no pointer sequence in front of it.
+    disownClicksUntil.current = Date.now() + CLICK_DISOWN_MS;
```
ewanhardingham, line 227: **The one deliberate spec deviation — worth your call.**

This used to fire only for a *classified* gesture (`if (active.classification !== null)`). It now fires for every completed gesture, taps included, because the recognizer answers the tap here and the synthetic click that follows would answer it a second time: re-revealing the pair a tap just concealed, or revealing on the first tap of a Check that #144 requires to cost nothing.

The price is that #138 §12's "Enter, Space **or a click** toggles reveal and conceal" is now only two-thirds true — a desktop mouse click on a face-down pair does nothing. Keyboard is unaffected (Enter/Space raise a click with no pointer sequence in front of it), the corner bend still works with a mouse, and a mouse double-click now Checks.

The alternative is branching on `event.pointerType` so mouse keeps its click and touch gets the recognizer — but then the two paths would have to agree about `TAPPED` as well, and the module has no other pointer-type branch. Happy to swap if you'd rather keep the click.

packages/player-client/src/holecards/useHoleCards.ts
```diff
@@ -156,28 +176,76 @@ export function useHoleCards(props: HoleCardPairProps): HoleCards {
...
+   * by the identical route and `canAct` — inside the intent, on the latest
+   * view — is the single gate deciding whether it may be sent. An off-turn or
+   * illegal double-tap therefore needs no guard here to be a no-op.
+   *
+   * `checkLegal` and `pending` are read for **rendering only**, exactly as the
+   * port promises: they decide whether to claim the Check landed. A stale prop
+   * can at worst show a confirmation for an Action the intent then refuses —
+   * never send one.
+   */
+  const sendCheck = (): void => {
```
ewanhardingham, line 208: Note what this function does *not* do: it never asks whether the Check is allowed. `props.actions.check` is `intent.check`, so `canAct` on the latest view decides, and an off-turn or illegal double-tap is inert without a guard here.

`confirmsCheck` reads `checkLegal`/`pending` purely to decide whether to *claim* it landed. Those two flags are derived in `Hand.cardActionsFrom` from the same `intent.legalActions`/`intent.pendingAction` that `canAct` consults, at the same instant — so the confirmation tracks the real outcome rather than guessing at it.

packages/player-client/src/holecards/cardState.ts
```diff
@@ -278,13 +278,34 @@ export function reduce(state: CardState, event: CardEvent): CardState {
       if (state.presentation === "Absent") return state;
       return idle("FaceDown");
 
+    case "TAPPED":
```
ewanhardingham, line 281: The load-bearing half of this arm is the *else*: a tap on a face-down pair does nothing. Revealing here would put a face-up frame between the two taps of a Check and charge the commonest free Action a reveal — which is exactly the acceptance criterion "face-down state: first tap is no-op, second sends Check".

packages/player-client/src/holecards/cardState.ts
```diff
@@ -278,13 +278,34 @@ export function reduce(state: CardState, event: CardEvent): CardState {
...
+      // tap would put a face-up frame between the two taps and charge the
+      // commonest free Action a reveal.
+      //
+      // Mid-turn it is dropped rather than queued, for the same reason
+      // `ACTIVATED` is: the flip is a point of no return.
+      return state.presentation === "Revealed"
+        ? { ...state, presentation: "FaceDown" }
+        : state;
+
+    case "DOUBLE_TAPPED":
```
ewanhardingham, line 294: Reducing an event that changes nothing looks like ceremony, but it buys the `survivesLock` guarantee for free: a decided (showdown) hand drops `DOUBLE_TAPPED` at the top of `reduce` without this arm having to remember. The Action itself is an effect, so the hook sends it.

packages/player-client/src/holecards/taps.ts
```diff
@@ -0,0 +1,62 @@
...
+ * Whether a double-tap Check should claim, visibly, that it landed (story 31).
+ *
+ * **Rendering only.** Whether the Action is *sent* is not decided here and not
+ * decided by these two flags: `check` is `intent.check`, and `canAct` on the
+ * latest view is the single gate (§2). This function answers the different
+ * question of whether to tell the player it went — so a double-tap off-turn,
+ * or while a Fold is in flight, is silent as well as inert, and a stale prop
+ * can at worst confirm an Action the intent then refuses.
+ */
+export function confirmsCheck(
```
ewanhardingham, line 58: **Judgement call I'd flag on the standards axis.** This predicate is a fact about `CardActions`, not about tap timing, so `ports.ts` is arguably its home and `taps.ts` now changes for two reasons.

I kept it here because it exists only to serve the double-tap path and reads as one thought with `tapLanded` — but it is the seam most likely to drift if the fold gesture (#145) grows a confirmation of its own.

packages/player-client/src/holecards/taps.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Answer a tap that has just landed: on its own, or as the second half of a
+ * double-tap.
+ *
+ * A completed pair closes the window rather than carrying the second tap
+ * forward, so three taps in quick succession send **one** Check and start
+ * counting again — a stream of taps cannot become a stream of Actions.
+ */
+export function tapLanded(window: TapWindow, now: number): TapStep {
```
ewanhardingham, line 41: Minor edge case, left alone deliberately: a *classified* gesture doesn't close the window, so tap → quick bend-and-release → tap, all inside 280ms, would still compose a Check. The middle gesture needs 9px of travel plus a release inside that budget, so I judged the rule not worth the extra state. Easy to add in `endGesture` if you'd rather it were airtight.

packages/player-client/src/holecards/coaching.ts
```diff
@@ -18,6 +18,13 @@ export interface Hint {
   readonly id: string;
   readonly line1: string;
   readonly line2: string | null;
+  /**
+   * Whether the hint is news rather than advice, and should therefore reach a
+   * screen reader when it appears. Only the Check confirmation is: every
+   * teaching hint and in-gesture prompt describes what the player *could* do,
+   * and announcing those over a live gesture would be noise.
+   */
+  readonly announce: boolean;
```
ewanhardingham, line 27: A boolean that is `false` on every hint but one — the other option was `hint.id === "checked"` at the render site, which couples rendering to a copy key. This way the copy table states its own semantics: the Check confirmation is news (`role="status"`), everything else is advice, and a live region over the in-gesture prompts would read a bend aloud as the finger crossed the diagonal.

packages/player-client/src/holecards/coaching.ts
```diff
@@ -55,6 +64,12 @@ export function selectHint(
   _discovered: ReadonlySet<TeachableGesture>,
   ctx: HintContext,
 ): Hint | null {
+  // Outranks every gate below, including the pending one it necessarily trips:
+  // this is not advice about a gesture, it is the answer to one the player just
+  // made (story 31). Without it the only sign a double-tap landed is the
+  // ActionBar the gesture exists to stop them watching.
+  if (ctx.checkConfirmed) return checkedHint;
```
ewanhardingham, line 71: Placed above the `pending` gate on purpose: sending the Check is precisely what makes an Action pending, so a confirmation that deferred to that gate could never appear at all. There is a test pinning this — it is the kind of ordering that looks like an oversight and gets "fixed" later.

packages/player-client/src/holecards/taps.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    let window = tapLanded(null, 0).window;
+    for (const now of [400, 800, 1200]) {
+      const step = tapLanded(window, now);
+      expect(step.event).toEqual({ type: "TAPPED" });
+      window = step.window;
+    }
+  });
+});
+
+describe("confirmsCheck", () => {
```
ewanhardingham, line 59: This is as close as the repo can currently get to the acceptance criteria "off-turn double-tap is a no-op" and "double-tap when Check is illegal is a no-op": the *sending* side is inert by construction (`canAct` inside the intent), and this pins the visible side.

Worth noting the gap it implies — with `environment: "node"` and no jsdom, nothing tests that a double-tap actually calls `actions.check()`. #138's testing strategy names a component-level integration layer that this repo has never had. It didn't matter for #141/#143; this is the first slice where an Action leaves the module on a gesture, so it may be worth deciding separately.

ewanhardingham (commented):
Inline notes on the decisions worth a second opinion — the click-disowning trade-off is the one that genuinely needs your call.

packages/player-client/src/holecards/useHoleCards.ts
ewanhardingham: I am happy with this deviation. No action required.

packages/player-client/src/holecards/coaching.ts
ewanhardingham: Agreed on the design — the copy table stating its own semantics is the right seam, and `hint.id === "checked"` at the render site would have been worse. But the wiring did not deliver it, so I changed that in d606029.

`role="status"` was on the hint `<p>` itself, and `GestureHint` returns `null` when there is no hint. So the live region was being *inserted into the document already holding its text*. Assistive technology announces changes to regions it is already observing; a region that arrives together with its content is not reliably announced at all — which would have made the confirmation silent in exactly the case it exists for.

`Announcer` is now a visually-hidden `role="status"` span mounted for the pair's lifetime and empty until there is news, so the region is there before the text is. The visible copy takes `aria-hidden` when `announce` is set, so the confirmation is read once rather than twice. `announce` itself is unchanged and still the thing that drives it.

packages/player-client/src/holecards/taps.ts
ewanhardingham: Closed it in d606029, on the grounds that the failure mode is asymmetric: the cost of the rule is a boolean, and the cost of not having it is a Check the player never asked for. Sending an unrequested Action is the one failure on this surface that costs money, so it seemed worth more than the 9px-plus-a-release improbability of hitting it.

`finish` now clears `tapWindow` when a completed gesture landed no tap, so the two taps must be consecutive. Deliberately *not* in `taps.ts` — the pure module knows about window arithmetic, and "what a non-tap gesture does to the window" is a fact about how gestures are sequenced, which is the hook's job. `tapLanded` is untouched.

packages/player-client/src/holecards/taps.ts
ewanhardingham: Keeping it here. `ports.ts` defines the seam's types, and a rendering predicate would be the first *behaviour* in it — a worse precedent than the one you are worried about. The coupling is also weaker than it looks: `confirmsCheck` depends on two flags, not on `CardActions`, which is why the parameter is a `Pick`.

If #145 grows a fold confirmation, the thing worth extracting is the confirmation concept shared by both gestures, not this function relocated. Revisit then.

packages/player-client/src/holecards/taps.test.ts
ewanhardingham: The claim is accurate — `environment: "node"`, `renderToStaticMarkup` only, no jsdom — and nothing here proves a double-tap reaches `actions.check()`. Agreed it should be decided separately rather than smuggled into this PR: standing up a DOM test environment is a repo-wide testing-strategy change, and #138's strategy is the right place to answer it. Worth its own issue against that section.

One thing this PR did add on the render side: d606029 pins that the `role="status"` region is in the document and empty on a settled pair, which `renderToStaticMarkup` can see.

packages/player-client/src/holecards/useHoleCards.ts
ewanhardingham: Checks out. `check` is `intent.check`, `canAct` is the only gate, and `checkLegal`/`pending` are read only to decide whether to claim it landed — no action.

packages/player-client/src/holecards/cardState.ts
ewanhardingham: Agreed, and the else is the acceptance criterion — no action.

packages/player-client/src/holecards/cardState.ts
ewanhardingham: Agreed: the free `survivesLock` guarantee is worth the arm that appears to do nothing — no action.

packages/player-client/src/holecards/coaching.ts
ewanhardingham: Agreed, and the test pinning the ordering is what stops it being "fixed" later — no action.

PR #162, opened 2026-08-02
fix(table,player): align top-bar pills and lobby controls

ewanhardingham:
## Intent

This PR is a focused UI polish pass for the table and player interfaces. The goal is to make the top-bar status information fit cleanly within its bounds, keep controls in predictable places as the room moves between lobby and live-game states, and make the one clickable pill look like an action rather than passive status.

### Table interface

- Keeps the **Connected** status pill inside the top bar on the far right in every state.
- Places the **room-code** pill on the far left, with equal top and bottom padding.
- Hides the room-code pill while the QR join card is open; shows it only when the game starts and the QR card disappears.
- Shows **Deal hand** and **End session** horizontally below the QR card while the room is in the lobby.
- Keeps **Deal hand** visible but visually and functionally disabled until enough players are seated.
- Returns the lobby controls to the existing right-hand rail once the hand has started.
- Keeps the lobby controls on the same stacking level as the QR card and burger button so they are not obscured by the overlay.

### Player interface

- Gives the name, seat number, sit-out, and connection pills a shared size and geometry.
- Differentiates the **Sit out / Sit in** control with an accent action treatment, pointer cursor, and disabled styling so it is clear that it is clickable rather than an indicator.

## Reviewer focus

The intended behavior is state-dependent placement, not just a spacing adjustment:

1. In the lobby, the QR card is the primary room-join surface; the table top bar shows Connected on the right, while the room-code pill is hidden.
2. Once a hand starts, the QR card disappears and the room-code pill appears on the left; Connected remains on the right.
3. The lobby actions sit beneath the QR card and are usable at the same visual layer. Deal hand remains present but cannot be activated until the player-count requirement is met.
4. On the player interface, all top pills line up consistently, while Sit out / Sit in reads as an action.

## Validation

- Focused table-client tests passed.
- Focused player-client tests passed.
- Table and player type checks and app builds passed.
- Prettier and ESLint checks passed.

### DIFF
--- a/packages/player-client/src/SeatPanel.test.tsx
+++ b/packages/player-client/src/SeatPanel.test.tsx
@@ -1,3 +1,4 @@
+import { color } from "@table-top-poker/ui-shared";
 import React from "react";
 import { renderToStaticMarkup } from "react-dom/server";
 import { describe, expect, it } from "vitest";
@@ -113,6 +114,13 @@ describe("SeatPanel", () => {
     );
 
     expect(html).toMatch(/data-testid="sitting-out-toggle"[^>]*disabled/);
-    expect(html).toContain("background:rgba(255,255,255,.04)");
+
+    // The disabled fill comes from PillButton, not from SeatPanel — the accent
+    // treatment must not survive into the disabled state.
+    const buttonMatch =
+      /<button[^>]*data-testid="sitting-out-toggle"[^>]*>/.exec(html);
+    expect(buttonMatch?.[0]).toContain(`background:${color.controlFill}`);
+    expect(buttonMatch?.[0]).toContain(`color:${color.disabledText}`);
+    expect(buttonMatch?.[0]).not.toContain("cursor:pointer");
   });
 });
--- a/packages/player-client/src/SeatPanel.test.tsx
+++ b/packages/player-client/src/SeatPanel.test.tsx
@@ -3,6 +3,7 @@ import React from "react";
 import { renderToStaticMarkup } from "react-dom/server";
 import { describe, expect, it } from "vitest";
 import { SeatPanel } from "./SeatPanel.js";
+import { playerTopPillStyle } from "./topPillStyle.js";
 
 const noop = () => undefined;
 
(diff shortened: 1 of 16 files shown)


ewanhardingham:
### Keep in mind — non-blocking

Review points that I deliberately did **not** action in `b22c37c..9fca291`. None of them block the merge; noting them so they're not lost.

**Uppercase on a `PillButton` cuts against its own doc comment.** `playerTopPillStyle` sets `textTransform: uppercase` and `letterSpacing: 0.16em`, and it's spread into the sit-out `PillButton`. `PillButton`'s comment says the opposite: *"No `textTransform` and normal tracking: every pill reads in the sentence case its label is written in, so a rail of mixed tones doesn't look like two different vocabularies."* So "Sit out" renders as "SIT OUT" while every other pill stays sentence case. If matching the neighbouring status pills is the intent — it reads like it is — the comment in `PillButton` is now the thing that's wrong and wants a sentence about pills that sit in a row of status chips.

**Smaller things:**

- `playerTopPillStyle` is exported from `SeatPanel.tsx` and imported by `StatusBar.tsx`. The direction is fine (`StatusBar` already renders `SeatPanel`), but shared pill geometry is easier to find in `ui-shared`, or in a local `styles.ts`, than in a component module.
- `StatusBar.tsx` (table): `justifyContent: space-between` no longer does anything — `marginLeft: auto` on the badge covers both the room-code-shown and room-code-hidden cases. Worth keeping only one.
- `TableControls.tsx` rebuilds `layoutStyle` and `actionButtonStyle` on every render from constant values. Hoisting the two placement variants to module scope would match how the rest of these components are written. No measurable cost either way.
- `App.tsx`: `!handInProgress ? (<TableControls …/>) : undefined` reads more naturally as `handInProgress ? undefined : …`, or `{!handInProgress && …}` if `JoinPanel` tolerates `false`.
- Some tests assert on serialised CSS rather than behaviour — `padding:16px 22px`, `flex-direction:row`, and `expect(html.match(/height:30px/g)).toHaveLength(3)`, which counts one magic number across the whole document and won't say which element regressed. The ordering assertions alongside them are the ones carrying signal. I left these as-is since they're pre-existing style in this file, but the `height:30px` count is the one most likely to fail for an unrelated reason.

ewanhardingham:
Actioned the smaller things from the note above in 6fe16e2 — the shared pill style now has its own module, the redundant `justifyContent` is gone (with a comment on why the auto margin is the load-bearing one), the `TableControls` placement styles are hoisted, the `controls` ternary reads condition-first, and the `height:30px` count is now scoped per element.

Still open by choice: the uppercase `PillButton` label, which needs a call on whether the code or `PillButton`'s doc comment is the stale one.

Issue #171, opened 2026-08-07
bug(table-client): End session leaves the previous hand in state for the next room


ewanhardingham:
Found during #148 on the latest Raspberry Pi build with the iPad as Table and iPhone/Android Players. Reproduction: create a Room, connect Players, complete or observe a Hand so the Table has non-empty hand/result state, click End session, then immediately create another Room without refreshing. Actual: the join code changes but the Table still displays state from the previous Room/Hand. Expected: ending a Room and creating another clears the prior Hand, showdown, board, seat/session presentation before the new Room receives its own state. Impact: a new Room shows stale information from the previous Room and can contaminate acceptance testing.

Issue #148, opened 2026-08-01
test(player-client): Phase 3 real-device acceptance run


ewanhardingham:
## Parent

Part of #138.

## What to build

Automated tests do not establish that this works. This ticket is the manual acceptance run that does.

**Emulation alone is insufficient.** Acceptance requires at least **one real iPhone/Safari device**, **one real Android/Chrome device**, and a **desktop mouse/keyboard pass**.

The iPhone/Safari pass carries particular weight: browser vibration is Blink-only, so on that device the permanent in-gesture text is the *only* arming signal for the one irreversible, money-losing gesture.

## Acceptance criteria

The manual matrix, on each required device:

- [x] Peel and release
- [x] Peel past the reveal threshold
- [x] Tap to conceal
- [x] Double-tap Check, on turn and off turn
- [x] Fold below the threshold
- [x] Fold above the threshold
- [x] Fold rejection return (interrupts the flight rather than completing it)
- [x] Pointer cancellation mid-gesture
- [x] Backgrounding mid-gesture
- [x] A server view arriving mid-peek
- [x] Showdown auto-reveal
- [x] Hint appearance and retirement across a **fresh device profile**
- [x] Long-press raises no callout menu
- [x] Double-tap causes no zoom
- [x] iPhone/Safari: the fold arming signal is legible with no haptic
- [x] Desktop: the whole surface works with a mouse, and the pair is reachable and toggleable by keyboard

Findings are recorded on this issue. Anything that fails becomes its own ticket rather than being fixed silently here.

## Conflict notes

None — no code changes expected. Orientation change mid-Hand is **out of scope** per the spec and is not a blocker; it graduates to its own ticket only if it proves to matter on a real device.

## Blocked by

- #146 — disarm the fold drag when the turn is lost.
- #147 — coaching hints that teach and retire.

ewanhardingham:
Acceptance run — iPhone/Safari: Peel and release — PASS. Short bottom-right peel showed the cards temporarily and release returned them face-down without an action or table change.

ewanhardingham:
Acceptance run — iPhone/Safari: Peel past reveal threshold — PASS. Dragging beyond the threshold left both cards face-up and persistent after release, with no table/action change.

ewanhardingham:
Acceptance run — iPhone/Safari: Tap to conceal — PASS. One tap hid both revealed cards immediately with no action.

ewanhardingham:
[...]

Issue #147, opened 2026-08-01
feat(player-client): coaching hints that teach and retire


ewanhardingham:
## Parent

Part of #138.

## What to build

On a touch device, the phone teaches itself. A small hint near the cards names **the one gesture the Player has not found yet** — in the order **Bend → Conceal → Check → Fold** — and each retires permanently once that gesture is used. A Player on a laptop or keyboard sees none of it.

This layers the teaching hints on top of the in-gesture prompts already shipped with #141 and #145, extending the same module-internal `selectHint`.

- `selectHint` is **module-internal** (§11). #111 placed hint discovery outside the card module; #135 then made every in-gesture hint depend on recognizer state nothing outside may see. The resolution: only the persisted discovery set crosses a seam, and the module's public surface stays at two names.
- `hintStorage.ts` loads and saves the discovery set through an **injected `Storage`**, exactly as `storage/seatToken.ts` does, so tests pass a double rather than reaching for a global. Persistence is per-device `localStorage`, permanent, never re-shown.
- **Touch-only**, via `window.matchMedia("(pointer: coarse)")` **subscribed to `change`**, not read once — a tablet that gains or loses a trackpad follows the current pointer. `any-pointer: coarse` is rejected: it is true for any touchscreen laptop, exactly the case the gate excludes.
- **Four teaching hints, not five.** Bend is Peek and bend past 0.9 is Reveal — one gesture at two depths — so the bend hint carries both outcomes on its second line.
- **At most one hint visible**, ordered: in-gesture always wins → Bend → Conceal (only while `Revealed`) → Check (only when legal) → Fold (only when legal). Bend and Conceal are eligible **off-turn**.
- **Timing:** a hint waits ~2s of no contact with the pair after its eligibility becomes true; it **hides the instant a pointer lands**; if still undiscovered it **returns after another ~2s of quiet**, so an aborted half-touch does not buy silence for the rest of the Hand.
- **Suppressed entirely** while an Action is pending, at Showdown, and when the pair is `Absent`.
- **Discovery is on-pair, never on a button.** Pressing Fold or Check in the `ActionBar` never counts — the hint exists to move the Player off the button.
- **"bottom-right" must track the rendered bend zone** rather than being hard-coded, so the copy follows the layout if it ever mirrors.

## Acceptance criteria

- [ ] At most one hint at a time; in-gesture outranks every teaching hint
- [ ] Bend outranks Conceal; Check outranks Fold in the overlap
[...]

PR #197, opened 2026-08-11
feat(player): land the coaching hints that teach and retire

### LINKED ISSUE #173: bug(player-client): Coaching hints absent in a fresh Android Chrome profile
Observed during #148 after opening a fresh Android Chrome private window and joining a newly created Room. With the Android Player on a live turn, the card pair was left untouched for more than 3 seconds. Actual: no coaching hint appeared. Expected: on a coarse-pointer fresh profile, the first Bend/peek-and-reveal hint appears after about 2 seconds of quiet. This may be downstream of stale Room/Hand state reported in #171 and #172; keep the observation, then re-test after those reset paths are fixed.

ewanhardingham:
Fixes #173.

## Why

#173 reports that on a fresh Android Chrome (coarse-pointer) profile, no Bend/peek coaching hint appears after ~2s of quiet. The root cause is structural: **the teaching hints never existed on `main`.**

The gesture *recognizers* and *in-gesture* prompts landed, but the four **teaching hints** (#147 — "Bend the corner…", "Double-tap to Check…", etc., which appear after the pair sits untouched) were built on the stale `feat/147-coaching-hints` branch and never merged. `main` carried only the seams:

- `HINT_QUIET_MS` defined but unused
- `selectHint` taking a `_discovered` set it ignored
- `HoleCardPair` hardcoding `quiet: false` and an empty discovery set

So "coaching hints absent" was true by design. #173's own hypothesis — that the original absence was downstream of the stale Room/Hand state in #171/#172 (both now fixed on `main`) — fits: the gating logic itself is sound; nothing was computing `quiet`/`coarsePointer`/`discovered`.

## What

Lands the teaching-hint layer on top of the existing scaffolding, adapting the `feat/147` work to `main`'s evolved `planFinish`-based finish path (a straight branch merge was not possible — `main` diverged substantially after `feat/147`).

- **`coaching.ts`** — four retiring teaching hints in teaching order (bend · conceal · check · fold), gated on a coarse primary pointer and the quiet interval; adds `nextTeachable` and `discoveredBy` as pure rules; `HintContext` gains `coarsePointer`.
- **`hintStorage.ts`** (new) — the per-device discovery set, persisted through an injected `Storage` (same pattern as `storage/seatToken.ts`); absent/malformed storage reads as nothing discovered.
- **`useHoleCards.ts`** — subscribes the `(pointer: coarse)` media query, times the quiet interval off contact with the pair, and routes the two pointer-sourced dispatch paths through a discovery wrapper so a gesture retires its own hint the first time it is used (never a button — discovery is on-pair by construction).
- **`HoleCardPair.tsx`** — builds the `HintContext` from the real `quiet`/`coarsePointer`/`discovered` values instead of the placeholder stubs; exports `HintBlock` for tests.
- **`constants.ts`** — `BEND_CORNER`, so the bend hint's copy tracks the rendered affordance rather than hard-coding "bottom-right".

## Tests

- `coaching.test.ts` extended: teaching-hint selection/ordering/eligibility, the coarse and quiet gates, `nextTeachable`, `discoveredBy`.
- `hintStorage.test.ts` (new): round-trip, absent/malformed reads, the allow-list, and a `setItem` that throws (Safari private mode).
- `HoleCardPair.test.tsx`: every hint's copy renders through the one `HintBlock`, and the taught corner matches the rendered bend zone.

Player-client suite: 349 passing. `tsc --build`, ESLint and Prettier all clean.

Note: player-client component tests render to static markup (node env), so the hook's quiet timer / `matchMedia` wiring is exercised through the pure `selectHint` unit tests rather than a live timer — the real-device confirmation belongs to the #148 acceptance run.

### DIFF
--- a/packages/player-client/src/holecards/HoleCardPair.test.tsx
+++ b/packages/player-client/src/holecards/HoleCardPair.test.tsx
@@ -2,7 +2,14 @@ import type { Card } from "@table-top-poker/protocol";
 import React from "react";
 import { renderToStaticMarkup } from "react-dom/server";
 import { describe, expect, it } from "vitest";
-import { HoleCardPair } from "./HoleCardPair.js";
+import type { CardState } from "./cardState.js";
+import {
+  selectHint,
+  type HintContext,
+  type TeachableGesture,
+} from "./coaching.js";
+import type { BendAxis } from "./geometry.js";
+import { HintBlock, HoleCardPair } from "./HoleCardPair.js";
 import type { CardActions } from "./ports.js";
 
 const actions: CardActions = {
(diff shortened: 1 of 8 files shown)


ewanhardingham:
Device acceptance passed on Android Chrome. Verified against the #173 gating on real hardware (the live quiet-timer + `(pointer: coarse)` wiring the node-env component tests can't exercise):

- **#173 core:** on a fresh coarse-pointer profile, `Bend the bottom-right corner` appears after ~2s of quiet, hides on contact, and returns after another quiet interval. ✅
- **Teaching sequence retires in order** — bend → conceal (face-up only) → check (when legal) → fold (when legal, taught last); Check ranked above Fold in the overlap. ✅
- **In-gesture prompts never retire** — bend (left/up second-line swap), fold (`Keep dragging up` → `Release to Fold`), and the `Checked` confirmation stamp. ✅
- **Negative gates** — no teaching hint mid-gesture, while pending, or at showdown; and a profile that has already bent does not see the bend hint again (persistence). ✅

Issue #160, opened 2026-08-02
Small blind and big blind markers on the table, and uniform marker sizing


ewanhardingham:
The table interface shows only the Button marker (`D`). Add Small Blind and Big
Blind markers alongside it, and fix the Button marker's size — all three markers
must render at the same size.

Outcome of a `/grilling` session; the decisions below are settled, not proposals.

## Acceptance criteria

1. In a hand with 3+ live players, the SB and BB seats each show a marker, in
   addition to the button's `D`.
2. SB/BB follow **ring order** from the button (SB = button+1, BB = button+2
   among live seats), never seat-number arithmetic.
3. **In a heads-up hand, only `D` is shown — no SB, no BB — throughout
   `betting`, `showdown` and `folded-out`.**
4. Between hands, only `D` is shown.
5. All three markers render at identical diameter and identical font size.
6. Colours: `D` off-white, `SB` blue, `BB` amber, dark ink on all three.
7. The markers all update on the same tick — the trio is never partially
   correct mid-hand.
8. Replay of an existing recorded log produces identical state;
   `ENGINE_LOG_VERSION` is unchanged.

### 1. Source of truth is the engine view

Add `smallBlind: SeatId` and `bigBlind: SeatId` to the hand-bearing view
variants, derived from `hand.ring` / `hand.button` via the existing
`bigBlindSeat` (`packages/engine/src/table.ts:70`) plus a new `smallBlindSeat`
sibling. The client renders what it is told.

Rejected: client-side derivation from `seats` + `button`. It would
re-implement both the heads-up rule and the live-seat filter, and would
disagree with the engine the moment someone sits out.

Consistent with the map's standing constraint that *"button and blinds are
tracked positionally, not by value"* — these are seat ids, no chips implied.

### 2. Markers move on the same tick, whenever a hand exists

All three derive from the same source as the `D` already rendered:
`hand.button` + `hand.ring`.

At `packages/engine/src/apply.ts:163-172` the engine button rotates on
`HandComplete`, but the completed `hand.button` keeps the old value and the
showdown/folded-out views render `hand.button` — so `D` only visibly jumps at
the `no-hand` transition. The blinds inherit exactly that timing.

**Between hands (`no-hand`), only `D` shows.** Blinds exist only while a hand
does. The engine button has already rotated by then, so a `no-hand` `D` is
already a forecast; extending that forecast to the blinds was rejected because
seats can be claimed, vacated or set to sit out before the deal, and the blinds
[...]

PR #206, opened 2026-08-15
PROTOTYPE (do not merge): player-side D/SB/BB position marker

ewanhardingham:
**Throwaway prototype — not for merge.** Opened as a pointer to the branch, per the prototype skill.

## The question

Where and how should the player's phone show that *they* are on the dealer button, small blind or big blind, in a style consistent with the markers the table already draws on its seat pods?

## Run it

```
npm run proto:position-marker
```

Six variants, switchable from the bar along the bottom; the URL carries `?variant=…&position=…`. The bottom bar also toggles `phase` and heads-up so the table's two marker-suppression rules can be checked, and prints the derived state on every switch.

| id | idea |
| --- | --- |
| `seat-pill-corner` | badge hung off the corner of the "Seat 3" pill — the table's exact move |
| `seat-pill-inline` | same disc, riding inside the pill |
| `top-row-chip` | its own pill in the top row: disc + the word |
| `banner-dot` | takes over the turn banner's status dot |
| `under-banner-strip` | a dedicated strip with the word and what it implies |
| `over-cards` | a physical-looking chip beside the hole cards |

## Findings so far

- **No protocol or server work is needed.** `PlayerView` already carries `button`, `smallBlind`, `bigBlind` and `dealtSeatCount` on every non-`no-hand` phase (`engine/src/view.ts`). This is purely a rendering decision.
- **Consistency is cheap.** The badge is lifted verbatim from `table-client/src/Seats.tsx` — same diameter/font-size split, same `color.buttonMarker`/`blindSmallMarker`/`blindBigMarker`, same `shadow.card` — so table and player show the same object.
- **One rule may not carry over.** The table suppresses SB/BB heads-up (#160, decision 4) to stop two markers landing on one seat. On the player screen you only ever see your own, so that rule would show a heads-up small blind nothing at all. Worth deciding explicitly.

## Scope

Additive only: one Vite entry, one `src/prototype-position-marker/` directory, one root script. No production component is touched, and `vite build` still emits only `index.html`, so the release is unaffected.

https://claude.ai/code/session_017UvUynydEB7FTmUauTDhV6

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "clean": "tsc --build tsconfig.json --clean",
     "test": "vitest run",
     "lint": "eslint . && prettier --check .",
+    "proto:position-marker": "npm run dev -w @table-top-poker/player-client -- --open /prototype-position-marker.html",
     "format": "prettier --write ."
   },
   "devDependencies": {
--- a/packages/player-client/prototype-position-marker.html
+++ b/packages/player-client/prototype-position-marker.html
@@ -1,0 +1,18 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1, viewport-fit=cover"
+    />
+    <title>PROTOTYPE — player position marker</title>
+  </head>
+  <body>
+    <div id="root"></div>
+    <script
+      type="module"
+      src="/src/prototype-position-marker/main.tsx"
+    ></script>
+  </body>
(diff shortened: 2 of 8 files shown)


ewanhardingham:
Verdict captured. The chosen placement is **`banner-dot`** — the marker takes the turn banner's tone-dot slot.

Implemented for real in #208 (issue #207). The decisions the prototype settled, plus the ones grilling surfaced that it hadn't covered, are recorded on #207.

This branch stays as the primary source and is **not for merge**.

Issue #178, opened 2026-08-09
Audio unlock & autoplay reality across the table and player clients


ewanhardingham:
Part of #177

## Question

Can each client reliably play a sound the moment a `HandEvent` arrives, given browser autoplay policy — and if not, what user gesture unlocks the AudioContext on each surface?

Verify against our actual targets:
- **table-client**: runs as a Pi-served kiosk on a TV/display. Does it ever get a user gesture? Candidates: room creation click, the settings toggle. If the page autostarts with no interaction, sound may be blocked until first tap.
- **player-client**: a phone browser. Join/seat-claim taps are gestures. Phones are also often on silent — note that reality.

Deliver: the concrete autoplay constraints (Chrome/Chromium kiosk + mobile Safari/Chrome), the recommended unlock gesture per surface, and any gotcha (suspended context after idle, muted-tab behaviour). AFK research — capture findings on a `research/audio-autoplay` branch and link from this ticket.

ewanhardingham:
## Resolution

Findings on the `research/audio-autoplay` branch (commit `1d06af3`), file `docs/research/audio-autoplay.md` — every claim cited to a primary source (Chrome for Developers, Chromium `media_switches.cc`, MDN, WebKit bugzilla, Apple dev forum).

- **Neither surface can play sound on a bare game event.** Chrome/Chromium and mobile Safari both create the `AudioContext` in `suspended` state when built before a user gesture, and keep it suspended until `resume()` (or a node `start()`) runs *inside* a gesture handler. Web Audio has been under Chrome's autoplay policy since Chrome 71.
- **Table kiosk gets no implicit gesture.** Cleanest fix: launch Chromium with `--autoplay-policy=no-user-gesture-required` (flag + siblings verified in Chromium source). Fallback: `resume()` on the room-creation click.
- **Phone unlock is free:** call `audioCtx.resume()` inside the existing join/seat-tap handler, which always precedes any sound-worthy event (MDN's create-or-resume-in-a-gesture pattern).
- **iOS silent-switch is the real gotcha:** Web Audio uses the "ambient" audio session and is muted by the hardware silent switch — unlocking the context does *not* make it audible on a silenced phone. Either accept that (arguably correct behaviour) or use the continuous silent-`<audio>` workaround to reroute onto the media channel.
- **Re-`resume()` on `visibilitychange`** — contexts can re-suspend when a tab is backgrounded/idle (matters for the phone; the kiosk is always foreground).

[...]

PR #229, opened 2026-08-16
fix(sound): recover iOS audio after another app takes the session

### LINKED ISSUE #228: bug(sound): iOS Safari never recovers audio after other media takes the session
## Symptom

On iPhone/Safari, sounds play normally after joining a game. If the player then plays other media — an Instagram video, a call, another tab — and comes back to the game tab, **the cues never come back**. The page stays silent for the rest of the session.

## Cause

iOS takes the audio session away from the page and parks the `AudioContext` in Safari's non-standard `"interrupted"` state, and pauses the silent keep-alive `<audio>` element that routes Web Audio through the media channel (#178). Neither undoes itself on return:

- the single `resume()` on `visibilitychange` in `webAudio.ts` is routinely ignored right after foregrounding, and there was no retry;
- an interruption Safari refuses to lift only clears with a *new* context — resume alone can never fix it;
- `play()` on the paused keep-alive element is rejected outside a user gesture, so the media-channel routing stays lost;
- every buffer source started against the parked context plays into silence, with nothing noticing.

## Fix

A recovery ladder in `webAudio.ts`, re-entered from every signal that the page is back (visibility, `pageshow`, `focus`, context `statechange`, and the next cue itself):

1. restart the keep-alive element and `resume()` the context, retrying with backoff;
2. if it is still not running and the page is visible, rebuild the `AudioContext` from scratch and re-decode the cues (rate-limited, and never for a context that has not yet had its first unlock);
3. otherwise wait for the user's next tap — capture-phase `pointerdown`/`touchend`/`click` listeners re-run the ladder, which is the one moment iOS reliably grants audio again.

Cues no longer start sources against a context that is not `running`.

ewanhardingham:
> **Status: does not fix #228.** Real-device testing shows sound still never returns after an interruption — and a page refresh or a Safari restart does not restore it either, which puts the cause outside anything this page can control. See https://github.com/ewanhardingham/table-top-poker/issues/228#issuecomment-5309496867. Held pending the diagnostics listed there.

Relates to #228 — on iPhone/Safari, playing other media (an Instagram video, a call, another tab) and returning to the game tab left the page silent for good.

## What was wrong

iOS parks the `AudioContext` in Safari's `"interrupted"` state and pauses the silent keep-alive element that routes Web Audio through the media channel (#178). Coming back undid neither: the single `resume()` on `visibilitychange` is routinely ignored right after foregrounding, an interruption Safari refuses to lift only clears with a *new* context, and every cue afterwards started a buffer source into silence with nothing noticing.

## What this does

A recovery ladder in `webAudio.ts`, re-entered from every signal that the page is back — `visibilitychange`, `pageshow`, `focus`, the context's own `statechange`, and the next cue itself:

1. restart the keep-alive element and `resume()` the context, retrying with backoff;
2. if it is still not running **and the page is visible**, rebuild the `AudioContext` and re-decode the cues — rate-limited to once per 5s, and never for a context still waiting on its first unlock gesture;
3. otherwise leave it: capture-phase `pointerdown`/`touchend`/`click` listeners re-run the ladder on the user's next tap, the one moment iOS reliably grants audio again.

`statechange` only takes step 1 — while the other app still holds the session a rebuild is wasted, so it is saved for the return. Cues no longer start sources against a context that is not `running`.

## Testing

New `webAudio.test.ts` fakes `AudioContext`/`<audio>`/the DOM events and covers the ladder's decisions: resume-on-return, no rebuild while the interruption is held, rebuild when resume stays refused, and recovery on the next tap when every automatic path failed. Three of the five fail against the old implementation.

`npm run lint`, `npm run build` and the ui-shared/player-client/table-client suites pass locally. The real-device check (interrupt with an Instagram video, return to the tab) still wants a human on an iPhone.

https://claude.ai/code/session_01FWVDggdkK43LhJqMudvKM3

### DIFF
--- a/packages/ui-shared/src/sound/webAudio.test.ts
+++ b/packages/ui-shared/src/sound/webAudio.test.ts
@@ -26,11 +26,20 @@ function eventTarget(): {
 
 class FakeAudioContext {
   static created: FakeAudioContext[] = [];
-  /** iOS-wide gate: false models a page with no audio permission right now. */
+  /** iOS-wide gate: false models the audio session being held by another app. */
   static audioAllowed = true;
+  /**
+   * True only inside the synchronous run of a tap handler. iOS grants audio to
+   * the task a gesture started, so a `resume()` issued after an `await` has no
+   * credit — modelling that is the point, since losing it is silence from the
+   * first hand rather than only after an interruption.
+   */
+  static inGesture = false;
   state: StateName = "suspended";
-  /** Set false to model Safari holding the interruption against `resume()`. */
-  resumable = true;
+  /** A context that has never run needs a gesture for its first resume. */
+  needsGesture = true;
+  /** Set true for the interruption Safari never lifts on this context. */
+  dead = false;
   /** Buffers handed to `start()` — the proof a cue actually made sound. */
   started: unknown[] = [];
   destination = {};
@@ -46,15 +55,18 @@ class FakeAudioContext {
     this.events.dispatch("statechange");
   }
 
-  /** Model the interruption: parked, and deaf to `resume()` until released. */
+  /** Model the interruption: parked, and deaf to `resume()` while held. */
   interrupt(): void {
-    this.resumable = false;
+    FakeAudioContext.audioAllowed = false;
     this.setState("interrupted");
   }
 
   resume(): Promise<void> {
-    if (!this.resumable || !FakeAudioContext.audioAllowed)
+    if (this.dead || !FakeAudioContext.audioAllowed)
       return Promise.reject(new Error("interrupted"));
+    if (this.needsGesture && !FakeAudioContext.inGesture)
+      return Promise.reject(new Error("gesture required"));
+    this.needsGesture = false;
     this.setState("running");
     return Promise.resolve();
   }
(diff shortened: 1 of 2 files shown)


ewanhardingham:
Regression fixed in d7a3389 — thanks for catching it, and sorry: my first commit broke unlock entirely.

**What I got wrong.** I routed `unlockAudio` through the new recovery ladder, and the ladder awaited the keep-alive element's `play()` *before* building the context and calling `resume()`. iOS grants audio to the task the tap started, so that await handed the credit back and the resume was refused — no sound at all, from the first hand, on iPhone and iPad alike. A `focus`/`pageshow` at load compounded it: it built a context outside any gesture and held the in-flight guard across the real unlock tap, so the tap skipped its own resume.

**What changed.**
- `unlockAudio` arms the keep-alive, builds the context and calls `resume()` synchronously, before anything is awaited.
- The ladder follows the same rule — the rebuild is synchronous now (the dead context is closed fire-and-forget) so a tap-driven run can rebuild *and* resume on that one tap.
- Passive recovery (visibility/pageshow/focus/gesture/statechange) does nothing until the app has asked to unlock.
- A resume refused while the interruption is still held no longer condemns the context; only one refused with the user back on the page does. Previously the first return always threw away a context that just needed resuming.

**Testing.** The fakes now model iOS's rule that audio belongs to the gesture's synchronous task, which is what the old tests missed. Six of the eight tests fail against the previous commit, including the plain "unlock from a tap" case; against `main` that case passes and the recovery cases fail, so the suite brackets the bug from both sides.

ewanhardingham:
Closing unmerged. The refresh/restart evidence on #228 shows the cause is outside anything this page controls, so this ladder is machinery for the wrong problem and `main` is better off without it.

The branch `fix/ios-audio-interruption-recovery` stays put — it has the recovery ladder and, more usefully, the first test harness for `webAudio.ts` (a fake `AudioContext` that models iOS's gesture-scoped audio permission). Worth cherry-picking from once the diagnostics on #228 say where the fix actually lives.

